PR #119, opened 2017-06-19
Experiment to build with Meson

jpakkane:
Hi

I was asked to submit this for consideration. This MR adds Meson build definitions to pgkconf. Some features:

 - tested to build and install on Linux
 - should work on Windows as well (including MSVC)
 - shared library versioning is missing
 - test suite is not run

Meson is being adopted by other low level Linux userland packages such as systemd, so it should be available in most scenarios.

### DIFF
--- a/libpkgconf/client.c
+++ b/libpkgconf/client.c
@@ -14,7 +14,7 @@
  */
 
 #include <libpkgconf/libpkgconf.h>
-
+#include <libpkgconf/config.h>
 /*
  * !doc
  *
--- a/libpkgconf/config.h.meson
+++ b/libpkgconf/config.h.meson
@@ -1,0 +1,88 @@
+/* libpkgconf/config.h.in.  Generated from configure.ac by autoheader.  */
+
+/* Define to 1 if you have the `cygwin_conv_path' function. */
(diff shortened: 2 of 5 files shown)


ignatenkobrain:
That was me who asked this, so I think it makes sense to replace cmake with meson part (and probably even autotools with meson).

kaniini:
@dankegel what do you think of this instead?

@jpakkane @ignatenkobrain because meson requires python, how does this change the bootstrap situation?  pkgconf is built quite early, much like systemd would be.

jpakkane:
> how does this change the bootstrap situation

For cross building not in any way as the build machine is going to have Python 3 (or it is simple to install). For native building the only requirement is that Python 3 and Ninja are invokable. FWICT systemd uses pkgconf so this might require building Python a bit earlier than currently (I'm not aware how early Python is built currently). I've been told that recent Python 3 versions are a lot easier to cross build than old versions were so this should be a fairly low effort change.

kaniini:
okay, so it seems like we can just cross-build pkgconf using meson, and that solves building a bootstrap image without having to worry about python (as python can be rebuilt on the host at a later point).

that makes sense, and it seems fine to me.

lets rip out cmake & autotools build systems and just use meson then.

PR #145, opened 2017-09-24
Only add warnings and std flags if supported by compiler (autoconf)

plicease:
This adds autoconf detection for the `-Wall` and `-std=` flags.  I am proposing as a separate PR as it adds a developer dependency on `autoconf-archive`.  If that is undesirable I can probably rework it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ pkg_default_dir        = @PKG_DEFAULT_PATH@
 pkgconfigdir           = $(libdir)/pkgconfig
 nodist_pkgconfig_DATA  = libpkgconf.pc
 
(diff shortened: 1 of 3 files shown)


kaniini:
Would it be possible to include the autoconf-archive macros somehow?

plicease:
Are you asking if the macro can be bundled with pkgconf?

kaniini:
yes, such as in `acinclude.m4`

plicease:
Rebased.

I bundled using AC_CONFIG_MACRO_DIRS as per this advice: https://www.gnu.org/software/automake/manual/html_node/Local-Macros.html
( if this is bad advice, please let me know :) )

I noticed that the bundled macro file includes license reference to GPL3 :( There is an "autoconf exception" for the license referred to here: https://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html .  I am not a lawyer, so I can't say with certainty if that would force a license change, which (seems to me) would be unacceptable.  My read is "autoconf exception" is fine for distributing the generated configure, but I am not sure about distributing just the m4 file.

PR #174, opened 2018-03-07
meson: Fix defining PKGCONF_API

TingPing:


kaniini:
thanks!  i would appreciate any sort of suggestions for CI of the meson build system with travis that we could cook up, too.

TingPing:
Should be pretty straightforward assuming `alpine` has it packaged.

kaniini:
we do :)

PR #180, opened 2018-04-03
libpkgconf: add support for Haiku

alaviss:
This patch add support for [Haiku]. The associated port can be found at https://github.com/haikuports/haikuports/pull/2026

[Haiku]: https://www.haiku-os.org/

### DIFF
--- a/libpkgconf/client.c
+++ b/libpkgconf/client.c
@@ -78,7 +78,11 @@ pkgconf_client_init(pkgconf_client_t *client, pkgconf_error_handler_func_t error
 	pkgconf_path_build_from_environ("PKG_CONFIG_SYSTEM_INCLUDE_PATH", SYSTEM_INCLUDEDIR, &client->filter_includedirs, false);
 
 	/* GCC uses these environment variables to define system include paths, so we should check them. */
+#ifdef __HAIKU__
+	pkgconf_path_build_from_environ("BELIBRARIES", NULL, &client->filter_libdirs, false);
+#else
 	pkgconf_path_build_from_environ("LIBRARY_PATH", NULL, &client->filter_libdirs, false);
+#endif
 	pkgconf_path_build_from_environ("CPATH", NULL, &client->filter_includedirs, false);
 	pkgconf_path_build_from_environ("C_INCLUDE_PATH", NULL, &client->filter_includedirs, false);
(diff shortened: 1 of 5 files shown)


libpkgconf/pkg.c
```diff
@@ -119,9 +119,28 @@ pkgconf_pkg_dir_list_build(pkgconf_client_t *client)
 
 	if (!(client->flags & PKGCONF_PKG_PKGF_ENV_ONLY))
 	{
+#ifdef __HAIKU__
```
kaniini: i'm not sure this is the way to go here.  should we maybe introduce more abstraction?

tests/regress.sh
```diff
@@ -79,7 +79,12 @@ variable_body()
 
 keep_system_libs_body()
 {
-	export PKG_CONFIG_PATH="${selfdir}/lib1" LIBRARY_PATH="/test/local/lib"
+	export PKG_CONFIG_PATH="${selfdir}/lib1"
+	if [ "$(uname -s)" = Haiku ]; then
```
kaniini: i don't like this approach, but since it's one test, I guess it's fine.  would prefer to handle Haiku in the same way as the other environments (see `test_env.sh.in`)

libpkgconf/client.c
```diff
@@ -78,7 +78,11 @@ pkgconf_client_init(pkgconf_client_t *client, pkgconf_error_handler_func_t error
 	pkgconf_path_build_from_environ("PKG_CONFIG_SYSTEM_INCLUDE_PATH", SYSTEM_INCLUDEDIR, &client->filter_includedirs, false);
 
 	/* GCC uses these environment variables to define system include paths, so we should check them. */
+#ifdef __HAIKU__
+	pkgconf_path_build_from_environ("BELIBRARIES", NULL, &client->filter_libdirs, false);
+#else
```
kaniini: you are certain GCC never checks `LIBRARY_PATH` on BeOS/Haiku?  that doesn't make much sense to me.  i would assume that `BELIBRARIES` is concatenated into `LIBRARY_PATH`, which means pkgconf_path_build_from_environ should be called on both env vars, in that order.

libpkgconf/pkg.c
kaniini: Yes, that would be fine.

libpkgconf/client.c
kaniini: OK.

PR #188, opened 2019-06-13
Fix libpkgconf visibility preset

evpobr:


a4z:
as the README.md says,  the project is no under there
https://git.dereferenced.org/pkgconf/pkgconf
and I think the github repo is unmaintained, or not synced. 
I found at least 1 thing that was fixed at git.dereferenced but not here.
Very unfortunate, this situation. 

So maybe want to check your changes with the code at git.dereferenced.org and make a RP there, if it is still required

PR #193, opened 2020-05-24
libpkgconf: personality: fix out of boundary access

stoeckmann:
It is possible to set the instruction pointer to undefined values by
using an operator larger than ':' in ASCII.

Since the personality function array does not have 256 entries, an
invalid operator can overflow the array.

Proof of concept:

$ echo "a _ b" > poc
$ ln -s $(which pkgconf) poc-pkgconf
$ ./poc-pkgconf

### DIFF
--- a/libpkgconf/personality.c
+++ b/libpkgconf/personality.c
@@ -168,7 +168,7 @@ personality_keyword_set(pkgconf_cross_personality_t *p, const size_t lineno, con
 	pair->func(p, keyword, lineno, pair->offset, value);
 }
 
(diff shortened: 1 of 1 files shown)


kaniini:
you seem to be having fun today :)

well I will put this in 1.7.1 and any other silliness you find :)

stoeckmann:
Heh, seems I am lucky to find these issues while inspecting the internal of pkgconf. Fortunately these issues are easy to fix, so I am fine with that approach.

kaniini:
This one has been merged.

PR #201, opened 2020-11-23
Add basic CI for mingw-w64 32/64 bit

lazka:
This only builds things with meson and no tests are run.

Not sure if this is something you want, feedback/questions welcome.

It should make reviewing Windows specific changes easier since you know at least that it builds fine :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,39 @@
+name: test
+
+on:
+  push:
+  pull_request:
+
+jobs:
+
+  msys2:
+    runs-on: windows-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


lazka:
Here is how it looks like: https://github.com/lazka/pkgconf/runs/1440848752

kaniini:
It would be nice to run the testsuite as well, but this is a good starting point.

lazka:
Thanks!

I can look into adding CI+tests using Linux if wanted.

kaniini:
The stuff in ifupdown-ng/ifupdown-ng should be sufficient for Linux CI.  I just haven't brought it over yet.

PR #202, opened 2020-11-23
Add CI for debian+meson/autotools

lazka:
Builds and runs tests. Using Debian testing since the test framework
is only in testing right now.


lazka:
Thanks!

PR #204, opened 2020-11-28
pkgconf_pkg_parser_version_func: fix whitespace detection

lazka:
In case the version string has no whitespace then strcspn() returns
strlen() of the input, so whitespace is only found if len != strlen.

This fixes invalid warnings when parsing version fields.


lazka:
This fixes all the warnings like:

```
> pkgconf.exe glib-2.0 --validate
/usr/lib/pkgconfig/glib-2.0.pc:12: warning: malformed version field with whitespace, trimming to [2.64.6]
```

lazka:
Thanks!

PR #226, opened 2021-06-23
Don't prepend sysroot_dir if pkg-config file lies outside of sysroot_dir

### LINKED ISSUE #213: PKG_CONFIG_SYSROOT_DIR get also prepended to libraries outside of SYSROOT
I've got a  problem when employing `PKG_CONFIG_SYSROOT_DIR`:

Not all of my libraries are contained in the sysroot, but there are other libraries built outside of the sysroot. (The sysroot is provided to me, so it is not so easily changed)

A concrete example:
my project needs libssh2, (which is not in the sysroot) is configured with `prefix=/home/tobi/libs/libssh2/usr `and it creates libssh2.pc with said prefix (libssh2.pc quoted in full at the bottom)

However, when `PKG_CONFIG_SYSROOT_DIR` is defined, for example to 
`PKG_CONFIG_SYSROOT=/home/tobi/sysroot`, the generated path 

`pkgconf --cflags libssh2`

will get me: 
`/home/tobi/sysroot/home/tobi/libs/libssh/usr/include`

which of course is not a valid path.

Should PKG_CONFIG_SYSROOT_DIR only be applied if the pc was coming from the sysroot?
 
TIA for any hints!

```
prefix=/home/EU.BSHG.COM/tobi/libs/libssh2/usr
exec_prefix=/home/EU.BSHG.COM/tobi/libs/libssh2/usr
libdir=${prefix}/lib
includedir=${prefix}/include

Name: libssh2
URL: https://www.libssh2.org/
Description: Library for SSH-based communication
Version: 1.9.0
Requires.private: libssl libcrypto zlib
Libs: -L${libdir} -lssh2
Libs.private:  -lgpg-error
Cflags: -I${includedir}
```

(note I've sent basically the same text also to pkg-config [1] and after I checked that pkgconf unfortunatly could not help me here, I thought it might be a good idea to post it here too.
On pkg-config. there is a PR pending with an another approach idea to us a new env map ( PKG_CONFIG_SYSROOT_MAP) to map pathes between several sysroots. Maybe interesting approach as well… [2] 

[1] https://gitlab.freedesktop.org/pkg-config/pkg-config/-/issues/52
[2] https://gitlab.freedesktop.org/pkg-config/pkg-config/-/merge_requests/7

Cheers,
tobi

manisandro:
Proposal to fix #213

Note: we hit this in Fedora, see [1]

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1974883

### DIFF
--- a/libpkgconf/pkg.c
+++ b/libpkgconf/pkg.c
@@ -406,6 +406,16 @@ pkgconf_pkg_new_from_file(pkgconf_client_t *client, const char *filename, FILE *
 	pkgconf_tuple_add(client, &pkg->vars, "pcfiledir", pc_filedir_value, true);
 	free(pc_filedir_value);
 
+	/* If pc_filedir is outside of sysroot_dir, clear pc_filedir
+	/* See https://github.com/pkgconf/pkgconf/issues/213
+	 */
+	if (client->sysroot_dir && strncmp(pkg->pc_filedir, client->sysroot_dir, strlen(client->sysroot_dir)))
+	{
+		free(client->sysroot_dir);
+		client->sysroot_dir = NULL;
+		pkgconf_client_set_sysroot_dir(client, NULL);
+	}
+
 	/* make module id */
 	if ((idptr = strrchr(pkg->filename, PKG_DIR_SEP_S)) != NULL)
 		idptr++;
--- a/tests/regress.sh
+++ b/tests/regress.sh
@@ -112,7 +112,7 @@ libs_never_mergeback_body()
 	export PKG_CONFIG_PATH="${selfdir}/lib1"
 	atf_check \
 		-o inline:"-L/test/bar/lib -lfoo1 \n" \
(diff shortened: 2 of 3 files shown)


Conan-Kudo (approved):

libpkgconf/pkg.c
```diff
@@ -406,6 +406,16 @@ pkgconf_pkg_new_from_file(pkgconf_client_t *client, const char *filename, FILE *
 	pkgconf_tuple_add(client, &pkg->vars, "pcfiledir", pc_filedir_value, true);
 	free(pc_filedir_value);
 
+        /* If pc_filedir is outside of sysroot_dir, clear pc_filedir
```
kaniini, line 409: Please fix the indentation to use actual tabs, thanks.

manisandro: Done

kaniini:
These changes are subtly broken...

PR #231, opened 2021-08-30
Fix typo

Olf0:
verses -> versus


Olf0:
#### Side note
It was fun reading the well written README, except for one paragraph:
> We will reject bugs like this, and if you choose to violate our Code of Conduct by demanding we fix your non-bug, we will fix the problem by banning you from the repo instead.

The tone might violate your CoC! ;-)
More seriously: You might want to phrase it a little less offensive, while still maintaining the message.  For example:

We will reject bugs like this, and if you choose to still demand fixing your non-bug, this constitutes a violation of our Code of Conduct, which might be addressed by banning you from this repository.

HTH

PR #248, opened 2022-04-01
Zero lists when freeing them

dcbaker:
Otherwise, the list pointer is left in a state that it cannot safely be re-initialized. This is particularly problematic for static members using lists, as once they are de-inited they cannot be re-iniited


kaniini:
Merged.

PR #256, opened 2022-07-29
Meson fixes and cleanups

dcbaker:
Initially I started on this because I'm using this in a subproject that set's it's C version to c99, and libpkgconf is full of warnings because of undefined functions. I realized that meson isn't setting a c_std unlike autotools. So i fixed that, added some warnings, updated the CI to be more picky, and then cleaned up the meson a bit (I am a meson dev, I might as well do something useful with all that knowledge :)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,8 +29,8 @@ check_functions = [
 ]
 
 foreach f : check_functions
-  if cc.has_function(f.get(1), prefix : '#define _BSD_SOURCE\n#include <' + f.get(2) + '>') and cc.has_header_symbol(f.get(2), f.get(1), prefix : '#define _BSD_SOURCE')
-    cdata.set(f.get(0), 1)
+  if cc.has_function(f[1], prefix : '#define _BSD_SOURCE\n#include <' + f[2] + '>') and cc.has_header_symbol(f[2], f[1], prefix : '#define _BSD_SOURCE')
+    cdata.set(f[0], 1)
   endif
 endforeach
 
--- a/meson.build
+++ b/meson.build
@@ -1,8 +1,9 @@
 project('pkgconf', 'c',
   version : '1.8.0',
   license : 'ISC',
-  meson_version : '>=0.47')
-
+  meson_version : '>=0.47',
+  default_options : ['c_std=c99'],
+)
 
 cc = meson.get_compiler('c')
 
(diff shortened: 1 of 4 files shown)


kaniini:
Looks fine, I will merge it into my tree today.  (If I click the button here, it will mess up when my personal git instance mirrors the repo back to Github)

dcbaker:
I updated this PR to add -Werror to the woodpecker CI as well, but otherwise this latest push is exactly the same as the last (since you mentioned it on the other PR)

meson.build
```diff
@@ -110,10 +119,8 @@ pkgconf_exe = executable('pkgconf',
 if get_option('tests')
   kyua_exe = find_program('kyua')
   atf_sh_exe = find_program('atf-sh')
-  test('kyua', kyua_exe, args : ['--config=none', 'test', '--kyuafile=' + join_paths(meson.build_root(), 'Kyuafile'), '--build-root=' + meson.build_root()])
-
-
-  configure_file(input : 'Kyuafile.in', output : 'Kyuafile', configuration : cdata)
+  kyuafile = configure_file(input : 'Kyuafile.in', output : 'Kyuafile', configuration : cdata)
```
tristan957, line 122: May want to match the `:` style with the rest of the files. Seem to be attached to the key elsewhere.

dcbaker: The style is all over the place here, and I'm inclined to leave it alone since all I did was assigned the output.

dcbaker:
Migrated to gitea

Tachi107:
> Migrated to gitea

This Gitea migration is kinda confusing to be honest. The README says to report bugs on GitHub, while the mailing list in on sourcehut, where there's also another pkgconf repo.

I guess that bug reporting link in the README simply has to be updated to point to Treehouse's Gitea instance (what's Treehouse btw? It's hard to determine by searching the web), but is the mailing list still monitored? What should it be used for?

dcbaker:
https://gitea.treehouse.systems/ariadne/pkgconf

Tachi107:
> https://gitea.treehouse.systems/ariadne/pkgconf is the url. We really should update the docs.

Yep, already forked that. My main question is about the sourcehut mailing list really (and also who/what Treehouse is)

PR #263, opened 2022-08-30
meson: remove useless command that isn't needed

eli-schwartz:
The tests/*.sh are executable in the source tree, and don't need to be chmodded after being copied to the build tree.


kaniini:
Merged, thanks

PR #266, opened 2022-09-30
Ignore whitespace indentation

### LINKED ISSUE #265: Should support indented .pc files
In the Debian bug #[1016922](https://bugs.debian.org/1016922) the reporter complains pkgconf cannot find spirv. Upon a closer look, I found that both spirv.pc and glslang.pc are indented by 8 spaces each.

Both pkg-config and ruby-pkg-config seem to be able these just fine; the spec as I understand it is silent on this.

It would probably make sense for pkgconf to support this as well.

Thanks for considering.

roehling:
This PR relaxes the parser to skip leading white space before a key and issue a warning. Currently, such lines are silently ignored, which differs from `pkg-config` behavior and causes surprising failures.

Fixes #265

### DIFF
--- a/libpkgconf/parser.c
+++ b/libpkgconf/parser.c
@@ -44,10 +44,18 @@ pkgconf_parser_parse(FILE *f, void *data, const pkgconf_parser_operand_func_t *o
 		lineno++;
 
 		p = readbuf;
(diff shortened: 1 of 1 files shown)


kaniini:
Merged, thanks

PR #270, opened 2022-10-13
pkg: make pkgconf_compare_version consistent

dwfreed:
The code taken from rpmvercmp in pkg-config returns -1 if a is less than 
b, 0 if a is equal to b, and 1 if a is greater than b. This matches the  
expectations of the comparison operators that use this function.         
However, the tilde handling, the NULL handling, and the docstring all do 
the opposite.                                                            
                                                                         
This fixes the tilde handling, the NULL handling, and the docstring to   
match the behavior of the rpmvercmp code and the expectations of the     
comparison operators.

The tilde handling fix is basically a revert of 22d7625

Before: https://godbolt.org/z/GfafdffY7

After: https://godbolt.org/z/Ysf7xjvTK

dpkg behavior for comparison:
```
$ dpkg --compare-versions '1.0.0~rc1' '<<' '1.0.0' && echo true || echo false
true
```

### DIFF
--- a/doc/libpkgconf-pkg.rst
+++ b/doc/libpkgconf-pkg.rst
@@ -76,7 +76,7 @@ routines.
 
    :param char* a: The first version to compare in the pair.
    :param char* b: The second version to compare in the pair.
-   :return: -1 if the first version is greater, 0 if both versions are equal, 1 if the second version is greater.
+   :return: -1 if the first version is less than, 0 if both versions are equal, 1 if the second version is less than.
    :rtype: int
 
 .. c:function:: pkgconf_pkg_t *pkgconf_builtin_pkg_get(const char *name)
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Additionally, as discussed at length in #mesonbuild, version comparison is currently a mess and should be properly defined.

No one that I know of anywhere has ever used tilde other than inside of a debian control file or rpm spec file, which doesn't overlap with pkg-config. Sort of makes sense, since tilde has basically never worked as intended in pkgconf.

It would probably make more sense to use a well defined scheme such as semver (1.2.3-rc1), or a natural scheme such as Python's PEP 440 (1.2.3rc1).

Freedesktop pkg-config doesn't support tilde, by the way.

kaniini:
Merged, thanks

PR #277, opened 2022-11-22
pkgconf:  Handle spaces correctly when expanding variables

zlaski:
Given the following .pc fragment:

includedir=/mingw64/include
Cflags: -I${includedir} -I${includedir}/taglib

Should includedir be assigned the value 'C:/Program\ Files/Git/mingw64/include', the expansion of ${includedir} will be chopped off after the first space:

Cflags: -IC:/Program\

With this patch, the expansion is corrected:

Cflags: -IC:/Program\ Files/Git/mingw64/include -IC:/Program\ Files/Git/mingw64/include/taglib

Create spaces-in-paths.pc


kaniini:
Merged, thanks

PR #280, opened 2023-01-12
Don't prepend sysroot_dir if pkg-config file lies outside of sysroot_dir

thorsten-klein:
Proposal to fix again issue https://github.com/pkgconf/pkgconf/issues/213 (the fix from Pull Request https://github.com/pkgconf/pkgconf/pull/226 is broken meanwhile)

### DIFF
--- a/libpkgconf/pkg.c
+++ b/libpkgconf/pkg.c
@@ -434,7 +434,9 @@ pkgconf_pkg_new_from_file(pkgconf_client_t *client, const char *filename, FILE *
 	 * See https://github.com/pkgconf/pkgconf/issues/213
 	 */
 	if (client->sysroot_dir && strncmp(pkg->pc_filedir, client->sysroot_dir, strlen(client->sysroot_dir)))
-		pkgconf_tuple_add(client, &pkg->vars, "pc_sysrootdir", "", false, pkg->flags);
+	{
+		pkgconf_client_set_sysroot_dir(client, "");
+	}
 
 	/* make module id */
 	if ((idptr = strrchr(pkg->filename, PKG_DIR_SEP_S)) != NULL)


kaniini:
I don't like this approach because it will break use of referenced dependent .pc files inside the sysroot.  More likely we should track and override sysroot_dir on per-pkg basis.  Does that make sense?

thorsten-klein:
Hello, I am completely open for any solution. The solution of this PR works fine for us. 
But if you have another solution, we are willing to try it as well.

kaniini:
Thinking about it, this is probably fine.  I will merge it into 2.0.1 release.

kaniini:
Your patch introduces regressions in the tests; I've made a note to look at this for the next release

kugel-:
Hello. I've came across this as well.

We have found that #226 *mostly* works for us. However, we discovered an edge case, when querying multiple modules, where it doesn't produce expected output.

module foo is within the sysroot, module bar is outside

`pkg-config --cflags foo` -> `-I<sysroot_dir>/usr/include`
`pkg-config --cflags bar` -> `-I/path/to/include`
`pkg-config --cflags foo bar` -> `-I<sysroot_dir>/usr/include -I/path/to/usr/include`
`pkg-config --cflags bar foo` -> `-I/path/to/usr/include -I/usr/include` :warning: 

In the last line, the second include path is not prefixed with sysroot_dir anymore.

Note: Since we use buildroot internally, I had to patch the elder pkgconf (1.6.x) within buildroot. For this I extracted the raw patch for #226 and added it to buildroot. It applies without issues but I might miss potential fixups.

kugel-:
Oh, as per your last comment, I did not try *this* patch, because a) it does not apply on-top of #226 and b)  I fear the regression.

dordnung:
So i run into the same problem and investigated the problem further. The problem why the current approach is not working anymore is, that it only clears the tuple “pc_sysrootdir” and not the “client->sysroot_dir” as in 1.8.0. But when parsing variables like “prefix”, “includedir” and so on, the “client->sysroot_dir” will be appended to the path in “pkgconf_tuple_parse” (https://github.com/pkgconf/pkgconf/blob/master/libpkgconf/pkg.c#L347) and (https://github.com/pkgconf/pkgconf/blob/master/libpkgconf/tuple.c#L318). This is why already the tuple “pcfiledir” (https://github.com/pkgconf/pkgconf/blob/master/libpkgconf/pkg.c#L429C41-L429C50) is incorrectly added. Maybe a better approach would be to use the flags of “pkg” to do not add the “client->sysroot_dir” to every path in the .pc file.

PR #281, opened 2023-01-17
Fix invalid dereferencing and comparing of empty strings.

Neumann-A:


cli/getopt_long.c
```diff
@@ -383,21 +383,23 @@ getopt_internal(int nargc, char * const *nargv, const char *options,
 	 * string begins with a '+'.
 	 */
 	posixly_correct = (getenv("POSIXLY_CORRECT") != NULL);
+	if(strcmp(options,"") != 0)
```
BillyONeal, line 386: Hmmm this seems equivalent to below, because the last character of an empty string is `'\0'`. Was this intended to be `options != 0` instead?

Neumann-A: that check is at https://github.com/pkgconf/pkgconf/blob/9aa08581976250549b2d0e085a7e51a7f970a0fe/cli/getopt_long.c#L378

> this seems equivalent to below

which part of below do you mean? 

Alternativly the code here could check `strlen(options)>0`

BillyONeal: > Alternativly the code here could check strlen(options)>0

If `strlen(options)` is `0`, then `*options == '-'` is valid (and `false`, because `*options` is `'\0'`). The only case this might do something different is if your CRT happens to make `strlen(0)` return `0`.

Neumann-A: >  then *options == '-' is valid (and false, because *options is '\0')

Just because you know an implementation detail of `const char*` doesn't make `*(vec.end())` a valid dereference. Especially in the case of `""` you might not even be the owner of the `\0` (since the compiler might point it to the end of any string in the read only section).  

The ASAN error is:
```
neumann@heineken MINGW64 /e/src/AllProjects/vcpkg/installed/x64-win-llvm-san-rel/tools/pkgconf (master)
$ ./pkgconf --version
=================================================================
==15024==ERROR: AddressSanitizer: odr-violation (0x7ff70c75eee0):
  [1] size=5 '"\0\0\0\0"' ../src/f90f789eb4-0a9c738510/cli/getopt_long.c
  [2] size=1 '""' ../src/f90f789eb4-0a9c738510/cli/main.c
These globals were registered at these points:
  [1]:
    #0 0x7ffe5c62654d  (E:\src\AllProjects\vcpkg\installed\x64-win-llvm-san-rel\tools\pkgconf\clang_rt.asan_dynamic-x86_64.dll+0x18003654d)
    #1 0x7ffee8a5e472  (C:\WINDOWS\System32\ucrtbase.dll+0x18001e472)
    #2 0x7ff70c751370 in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:256
    #3 0x7ffee9477613  (C:\WINDOWS\System32\KERNEL32.DLL+0x180017613)
    #4 0x7ffeead226a0  (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526a0)

  [2]:
    #0 0x7ffe5c62654d  (E:\src\AllProjects\vcpkg\installed\x64-win-llvm-san-rel\tools\pkgconf\clang_rt.asan_dynamic-x86_64.dll+0x18003654d)
    #1 0x7ffee8a5e472  (C:\WINDOWS\System32\ucrtbase.dll+0x18001e472)
    #2 0x7ff70c751370 in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:256
    #3 0x7ffee9477613  (C:\WINDOWS\System32\KERNEL32.DLL+0x180017613)
    #4 0x7ffeead226a0  (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526a0)

==15024==HINT: if you don't care about these errors you may set ASAN_OPTIONS=detect_odr_violation=0
SUMMARY: AddressSanitizer: odr-violation: global '"\0\0\0\0"' at ../src/f90f789eb4-0a9c738510/cli/getopt_long.c
==15024==ABORTING
```
The '"\0\0\0\0"' is https://github.com/pkgconf/pkgconf/blob/0226cdda6d6c764bf18df24869c043c85e11cdd1/cli/getopt_long.c#L89
(or probably more the place where the string/define is stored/used https://github.com/pkgconf/pkgconf/blob/0226cdda6d6c764bf18df24869c043c85e11cdd1/cli/getopt_long.c#L415)

The '""' is here https://github.com/pkgconf/pkgconf/blob/0226cdda6d6c764bf18df24869c043c85e11cdd1/cli/main.c#L922

If your replace EMSG with a `const char* EMSG`. ASAN will report a global buffer overflow at 
https://github.com/pkgconf/pkgconf/blob/0226cdda6d6c764bf18df24869c043c85e11cdd1/cli/getopt_long.c#L392
coming from 
https://github.com/pkgconf/pkgconf/blob/0226cdda6d6c764bf18df24869c043c85e11cdd1/cli/main.c#L922
(Apart from that having a `pkg_getopt_long_only` function which calls an internal function which checks long and short options and requires inputs for both is kind of strange since the function explicitly says `long_only`....)

If you want you can call it ASAN being overly correct.

BillyONeal: > Just because you know an implementation detail of `const char*` doesn't make `*(vec.end())` a valid dereference.

If `*(vec.end())` is done by `*options`, that is done just as much by `strlen(options)` or `strcmp(options)`. Hiding the dereference inside `strlen` or `strcmp` does not make the dereference any more valid.

> Especially in the case of "" you might not even be the owner of the \0 (since the compiler might point it to the end of any string in the read only section).

I'm not sure where ownership comes up here. This is a read, not a write.

>The ASAN error is:

If changing it to `strcmp` fixed the problem, that fixed it by hiding it from ASAN (because `strcmp` wasn't itself compiled with ASAN).

> ```
> ==15024==ERROR: AddressSanitizer: odr-violation (0x7ff70c75eee0):
>  [1] size=5 '"\0\0\0\0"' ../src/f90f789eb4-0a9c738510/cli/getopt_long.c
>  [2] size=1 '""' ../src/f90f789eb4-0a9c738510/cli/main.c
> ```

If ASAN is saying "these two totally unrelated literals are an ODR violation because the compiler internally decided to do string pooling", that sounds like a bug in ASAN (or the compiler), not a bug in pkgconf.

> If your replace EMSG with a const char* EMSG. ASAN will report a global buffer overflow at

If there is in fact a buffer read overflow, changing `*options` to `strlen(options)` did not fix it.

> If you want you can call it ASAN being overly correct.

I have not reproed the problem happening in ASAN. Whether ASAN is correct or not, this change didn't change the situation. Doing the lvalue-to-rvalue conversion on `*options` and calling `strlen(*options)` read `*options`.

Neumann-A: The code has `""` so it owns an empty string with a size of 0 bytes.
The C implementation behind it has to add some overhead (in this case 1 byte, '\0') so that it knows about empty strings and can differentiate between emptiness, nothingness and separate strings in memory. It could theoretically add more, e.g. "`<something_at_the_start><string_in_the_code><something_at_the_end>`" but the point here is that the code only owns `<string_in_the_code>` and should not access anything else. However in the case of an empty string the code access `<something_at_the_end>` which is an out of bound memory read from the codes perspective. Doing that read via C library functions however is ok since the C library owns all of `<something_at_the_start><string_in_the_code><something_at_the_end>` since it has to define how to encode char * in memory together with the compiler.

So doing `strlen(options)` is not hiding the issue it is delegating the memory access to the correct place. 

So from my point ASAN is just being pedantic. I wouldn't consider it a bug in ASAN since it makes sense to report something in this case. Is it a security/serious issue? Very probably not. Should pkgconf check for empty strings? Yes. Should pkgconf even have code for not used short options? probably no. 

> I'm not sure where ownership comes up here. This is a read, not a write.

So reading a private member via an offset from the this pointer without having a getter for you is ok? (e.g. `*(this+<some_magic_number_obtained_from_knowing_the_impl_defined_memory_layout>)` )

kaniini: The getopt_long code is just cribbed from NetBSD with a couple of tweaks.

BillyONeal: > The code has `""` so it owns an empty string with a size of 0 bytes.

No. `sizeof("")` is 1. https://gcc.godbolt.org/z/GGKT38oc4

> the point here is that the code only owns <string_in_the_code> and should not access anything else.

It also owns the null terminator.

> Doing that read via C library functions however is ok since the C library owns all of <something_at_the_start><string_in_the_code><something_at_the_end> since it has to define how to encode char * in memory together with the compiler.

That's just... not accurate. The way strlen works is very well defined and very much requires that there is a readable terminal null. From ANSI ISO IEC 9899 2011 "The C Standard":

image

There is no collusion between the standard library and what happens with literals. They are terminated by null just like anything else. If the program were trying to *modify* that null, then yes, that would be out of contract, but this is a read.

> So doing `strlen(options)` is not hiding the issue it is delegating the memory access to the correct place.

Nope.

ASAN isn't even reporting an out of bounds memory read in your example, it's reporting that 2 different literals in the program got interned together. Again, the only reason it isn't reporting the exact same error inside `strlen` is because `strlen` was not compiled with ASAN.

> So reading a private member via an offset from the this pointer without having a getter for you is ok? (e.g. `*(this+<some_magic_number_obtained_from_knowing_the_impl_defined_memory_layout>)` )

No, but that's nothing like what this is doing. Reading the null is in contract. http://eel.is/c++draft/lex.string#10 in C++ and 6.4.5 "String literals"/6 in C:
image


Accessing some random offset from a this pointer is out of contract. Reading the null is in contract. ASAN's output you have posted even agrees with this; it's complaining about an ODR violation, not an out of bounds read. (And an ODR violation the compiler committed no less)

Neumann-A: > No. sizeof("") is 1. https://gcc.godbolt.org/z/GGKT38oc4

That is measuring `<size_of_string>+<size_of_required_overhead>`. So you don't seem to understand what I am arguing about

> It also owns the null terminator.

No the program owns it. The code in question just owns the stuff between the `""` which is nothing. 

> From ANSI ISO IEC 9899 2011 "The C Standard":

Citing the standard is basically looking at the implementation. Basically Point 3 is something a user doesn't need to know, it is an implementation detail. 

> Accessing some random offset from a this pointer is out of contract. 

Why? Following your logic I just need to write a standard about it which defines the offset and then I am able to read it within contract. Also I am not saying there is any breach of contract. I am just saying that knowing nothing on how char* is implemented the code is accessing a byte it should strictly speaking not know anything about since it is an implicitly added byte by the standard. So it is a logically argument about the code not about the fact that we all know how char* is implemented and that reading the implementation defined and standardized null terminator will be totally fine. 

Also here is the error after removing `EMSG`, hardcoding the string and moving the place variable as a local static into the functions using it:
```
=================================================================
==15232==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7ff71cafea20 at pc 0x7ff71cad93a8 bp 0x005968f6efd0 sp 0x005968f6f018
READ of size 1 at 0x7ff71cafea20 thread T0
    #0 0x7ff71cad93a7 in getopt_internal E:\vcpkg_folders\llvm_test\buildtrees\pkgconf\src\6aa314c9c4-8bd74a463d\cli\getopt_long.c:391 [this is "if (posixly_correct || *options == '+')"]
    #1 0x7ff71cad98ba in pkg_getopt_long_only E:\vcpkg_folders\llvm_test\buildtrees\pkgconf\src\6aa314c9c4-8bd74a463d\cli\getopt_long.c:640 [this is "return (getopt_internal(nargc, nargv, options, long_options, idx,"]
    #2 0x7ff71cad13c9 in main E:\vcpkg_folders\llvm_test\buildtrees\pkgconf\src\6aa314c9c4-8bd74a463d\cli\main.c:903
    #3 0x7ff71caf0c5b in invoke_main D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:78
    #4 0x7ff71caf0c5b in __scrt_common_main_seh D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
    #5 0x7ff9410d7613  (C:\WINDOWS\System32\KERNEL32.DLL+0x180017613)
    #6 0x7ff9418e26a0  (C:\WINDOWS\SYSTEM32\ntdll.dll+0x1800526a0)

0x7ff71cafea20 is located 31 bytes to the right of global variable '"\"%s\" [fontname=Sans fontsize=8]\n"...' defined in '../src/6aa314c9c4-8bd74a463d/cli/main.c' (0x7ff71cafe9e0) of size 33
  '"\"%s\" [fontname=Sans fontsize=8]\n"...' is ascii string '"%s" [fontname=Sans fontsize=8]
'
0x7ff71cafea20 is located 0 bytes inside of global variable '""' defined in '../src/6aa314c9c4-8bd74a463d/cli/main.c' (0x7ff71cafea20) of size 1
  '""' is ascii string ''
SUMMARY: AddressSanitizer: global-buffer-overflow E:\vcpkg_folders\llvm_test\buildtrees\pkgconf\src\6aa314c9c4-8bd74a463d\cli\getopt_long.c:391 in getopt_internal 
Shadow bytes around the buggy address:
  0x1177a63dfcf0: 00 01 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00 00
  0x1177a63dfd00: 00 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00 00
  0x1177a63dfd10: 06 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 04 f9 f9 f9
  0x1177a63dfd20: f9 f9 f9 f9 00 00 00 00 02 f9 f9 f9 f9 f9 f9 f9
  0x1177a63dfd30: 00 00 00 00 01 f9 f9 f9 f9 f9 f9 f9 00 00 00 00
=>0x1177a63dfd40: 01 f9 f9 f9[f9]f9 f9 f9 00 00 00 00 00 00 00 00
  0x1177a63dfd50: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1177a63dfd60: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1177a63dfd70: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1177a63dfd80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x1177a63dfd90: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==15232==ABORTING
```

N-R-K: > I just need to write a standard about it which defines the offset and then I am able to read it within contract

Ah yes, just write your own standard so you can do whatever you want! Why did no one else think about this world changing idea which would easily solve all the bugs in the world?

kaniini: > Also here is the error after removing EMSG, hardcoding the string and moving the place variable as a local static into the functions using it

In other words, here is an error in code that you have modified to do something other than what the code presently does.

kaniini:
Anyway, as explained by @BillyONeal this isn't really a bug, just ASan being pedantic and applying C++ practices to a C project.  There is no "one definition rule" in C.

cli/getopt_long.c
BillyONeal: > That is measuring `<size_of_string>+<size_of_required_overhead>`. So you don't seem to understand what I am arguing about

The "required overhead" is something the program owns.

> No the program owns it. The code in question just owns the stuff between the `""` which is nothing.

Exactly, the program owns it. `""[0]` is entirely valid. (And is identical to `'\0'`)

> Citing the standard is basically looking at the implementation. Basically Point 3 is something a user doesn't need to know, it is an implementation detail.

The standard specifies only the interface, not implementation details.

> I just need to write a standard about it which defines the offset and then I am able to read it within contract. 

If the standard defined such an offset then yes, it would be in contract. But the standard does not. It does specify what string literals are.

> I am just saying that knowing nothing on how char* is implemented the code is accessing a byte it should strictly speaking not know anything about since it is an implicitly added byte by the standard.

If the documented behavior is that the char* points at a null terminated string, then reading the null terminator is in contact. `f("")` is in contract for a function documented to accent an NTBS. `f(nullptr)` would be out of contract.

The additional byte is part of the string.

>So it is a logically argument about the code not about the fact that we all know how char* is implemented

This has nothing to do with `char*`. This is about the type of `""` being `const char[1]`. Yes, you can form a pointer to that array and access anything in that array, including the terminator.

>Also here is the error after removing EMSG, hardcoding the string and moving the place variable as a local static into the functions using it:

If you replaced a literal like "\0\0\0\0" with a plain `const char*` that was initialized to nullptr then your edit was out of contract. The function is documented to accept a null terminated string, for which `const char* some_variable;` is not in contract, because you can't dereference `some_variable[0]`. If you did `const char* some_variable = "\0\0\0\0";` or `const char* some_variable = "";` it would be in contract again.

I don't know exactly what edit you made here to discuss what ASAN is reporting in that case, I can only speak confidently about the code in front of me.

BillyONeal:
> Anyway, as explained by @BillyONeal this isn't really a bug, just ASan being pedantic and applying C++ practices to a C project. There is no "one definition rule" in C.

There absolutely is a one definition rule in C:

image

I just don't think it applies in this example, because the entities ASAN is complaining about are literals (which have no linkage).

cli/getopt_long.c
Neumann-A: > The "required overhead" is something the program owns.

Yes of course. The question is rather if the code you write also owns it or should treat the type as a black box knowing only what was explicitly written down in the code. The program is always `<the current code>+<3rd party libs>+<c standard> `

> The standard specifies only the interface, not implementation details.
> It does specify what string literals are.

specifying a type which is guaranteed to be '\0' terminated is an implementation detail. Specific functions to ask relevant question for that type is an interface.  As such nobody outside the c std lib needs to know that the type char * will be null terminated.

In the end I am just saying that `strcmp(options,"") != 0` is better than `*options='\0'` because the former doesn't require extra internal knowledge how `char*` is structured. It just requires knowledge about the functions acting on the type defined in the c standard library. I think the reason '\0' is mentioned is for people actually implementing the c standard library and not having to come up with their own termination encoding. In the end the standard is made by humans and it is very hard to get people to agree to something as demonstrated here ;) 

> If you replaced ....

function local `static const char* place = "\0\0\0\0";` + string replacement of EMSG with "\0\0\0\0" (what the preprocessor would have done).

BillyONeal: >In the end I am just saying that strcmp(options,"") != 0 is better than *options='\0' because the former doesn't require extra internal knowledge how char* is structured. 

It requires the exact same knowledge about how the char* is structured. `strcmp(nullptr, "")` is out of contract, for the same reason. `strcmp` is documented to require pointers to NTBSes, so `*arg` must be valid.

Again, the only reason ASAN is not reporting the exact same error is because `strcmp` was written in assembly and thus not compiled with ASAN's checks, not because it is doing anything differently.

> I think the reason '\0' is mentioned is for people actually implementing the c standard library and not having to come up with their own termination encoding. 

It's mentioned that way because users have to be able to form their own NTBSes. Example:

```
char arr[4];
arr[0] = 'a';
arr[1] = 'b';
arr[2] = 'c';
arr[3] = 'd';
strcmp("hello", arr);
```

is out of contract, but

```
char arr[4];
arr[0] = 'a';
arr[1] = 'b';
arr[2] = 'c';
arr[3] = '\0';
strcmp("hello", arr);
```

is in contract.

>In the end the standard is made by humans and it is very hard to get people to agree to something as demonstrated here ;)

Are you seriously trying to claim that "null terminated byte string" is allowed to have some kind of special extra termination that is not a null? I am very confused.

>function local static const char* place = "\0\0\0\0"; + string replacement of EMSG with "\0\0\0\0" (what the preprocessor would have done).

Looking at this error message again:

> 0x7ff71cafea20 is located **31 bytes to the right** of global variable '"\\"%s\\" [fontname=Sans fontsize=8]\\n"...' defined in '../src/6aa314c9c4-8bd74a463d/cli/main.c' (0x7ff71cafe9e0) of size 33
>   '"\"%s\" [fontname=Sans fontsize=8]\n"...' is ascii string '"%s" [fontname=Sans fontsize=8]

31 bytes to the right certainly doesn't sound like an off by one read off the end of a string. It looks an awful lot like again, string interning done by the compiler without informing ASAN it did that.

BillyONeal: Please make sure that you aren't compiling with [`/GF`](https://learn.microsoft.com/en-us/cpp/build/reference/gf-eliminate-duplicate-strings?view=msvc-170)

Neumann-A: > Please make sure that you aren't compiling with [/GF](https://learn.microsoft.com/en-us/cpp/build/reference/gf-eliminate-duplicate-strings?view=msvc-170)

That might be a good hint:
`/GF                     Enable string pooling (default)`

And with `/GF-` ASAN is happy. (Maybe clang-cl ASAN should auto deactivated this)

> form their own NTBSes. Example:

I see. That is exactly the reason why c string functions are considered unsafe.  Why exactly do users need to define their own terminated char arrays which are able to correctly interact with string functions `strcmp` ? (use `const char arr[]= "abc";` instead)

> Are you seriously trying to claim that "null terminated byte string" is allowed to have some kind of special extra termination that is not a null? I am very confused.

Yeah you could do that. E.g. `\0Term\0` , `\0DontLookHere`. `T\3rminat\0r` could also be a valid termination encoding although not with the current standard. As long as the corresponding `str*` functions still fulfill their original meaning everything is ok. The reason that `\0` was chosen as the termination character is just because it had the smallest overhead and made kind of sense and thus a standard was written for exactly that. But that discussion is more philosophical than practical. 

> strcmp(nullptr, "") is out of contract

`nullptr` needs to be guarded but I never talked about that case since it is boring ;)

N-R-K: > This is about the type of `""` being `const char[1]`

Close enough, but string literals aren't `const` in C, although writing to them is UB. [^0] The `-Wwrite-strings` warning in gcc (and clang) is helpful in this regard.

- - -

[^0]: http://port70.net/~nsz/c/c99/n1256.html#6.4.5p6

>specifying a type which is guaranteed to be '\0' terminated is an implementation detail.

No, it's not an implementation detail, it's the *defining* feature of a string in C.

> It just requires knowledge about the functions acting on the type defined in the c standard library

Strings in C don't have a distinct type and `char` is not a type that's defined by the standard library - it's one of the basic types [^1].

[^1]: http://port70.net/~nsz/c/c99/n1256.html#6.2.5p14

If you want to learn how things actually work, the draft of the standard is [freely available online](http://port70.net/~nsz/c) to be read, which I recommend doing instead of making things up.

kaniini:
Sorry, I was tired.  What I mean is what you said, that the ODR violation allegedly happening here is not really an ODR violation, because literals are not subject to them.

At any rate, I don't care if vcpkg continues to carry the patch, but the patch seems dubious at best, so it is probably worth reverting.

Issue #287, opened 2023-02-28
Edge case in single quote escaping


lu-zero:
Here the testcase

escape.pc
```
Name: Escape
Version: 4.2.0
Description: Escape utility library
Libs: -Llink\ path\ with\ spaces
Cflags: -DA=\"escaped\ string\'\ literal\"
```

Here the different handling:

pkg-config escape --cflags
```
-DA=\"escaped\ string\'\ literal\"
```

pkgconf escape --cflags
```
-DA=\"escaped\ string\\\'\ literal\"
```

kaniini:
Fixed in pending 1.9.5 release, thanks!

PR #289, opened 2023-03-06
argvsplit: fix some quoting rules

huyubiao:
- Fix the issue：#287 
I built some test cases, which can cover all conditions, and it can be successfully executed under my Linux.


kaniini:
Applied, thanks!

Issue #288, opened 2023-03-06
tests/meson.build is deleted when make clean is executed


huyubiao:
Hello

I did `make clean` by accident. When I tried to execute `make check` again, I found that `tests/meson.build` had been deleted.
- Here's some output to help me spot where meson.build was deleted：
```
# make clean
# rm -f pkgconf bomtool
# test -z " tests/meson.build tests/basic tests/builtins tests/conflicts tests/framework tests/parser tests/provides tests/regress tests/requires tests/sysroot tests/version" || rm -f  tests/meson.build tests/basic tests/builtins tests/conflicts tests/framework tests/parser tests/provides tests/regress tests/requires tests/sysroot tests/version
test -z "libpkgconf.la" || rm -f libpkgconf.la
```
- Makefile.am:
```
test_scripts=	tests/meson.build \
		tests/basic.sh \
		tests/builtins.sh \
		tests/conflicts.sh \
		tests/framework.sh \
		tests/parser.sh \
		tests/provides.sh \
		tests/regress.sh \
		tests/requires.sh \
		tests/sysroot.sh \
		tests/version.sh

check_SCRIPTS=	$(test_scripts:.sh=)
CLEANFILES =	$(EXTRA_PROGRAMS) \
		$(check_SCRIPTS)
```

PR #290, opened 2023-03-07
Fix the issue where tests/meson.build was deleted.

huyubiao:
- https://github.com/pkgconf/pkgconf/issues/288
I removed `tests/meson.build` from variable `check_SCRIPTS`.


kaniini:
Applied, thanks!

Issue #291, opened 2023-03-17
Misuse of ctype(3) functions/macros


riastradh:
The C standard definition of the <ctype.h> functions, such as isspace, isalpha, isdigit, &c., says:

> In all cases the argument is an int, the value of which shall be representable as an unsigned char or shall equal the value of the macro EOF.  If the argument has any other value, the behavior is undefined.

This is because they're designed to work with the int values returned by getc or fgetc; they need extra work to handle a char value, for instance when processing a string stored in a char array.

If EOF is -1 (as it almost always is), with 8-bit bytes, the allowed inputs to the ctype(3) functions are:

{-1, 0, 1, 2, 3, ..., 255}.

However, on platforms where char is signed, such as x86 with the usual ABI, code like

```
char *ptr = ...;
... isspace(*ptr) ...
```

may pass in values in the range:

{-128, -127, -126, ..., -2, -1, 0, 1, ..., 127}.

This has two problems:

1. Inputs in the set {-128, -127, -126, ..., -2} are forbidden.

2. The non-EOF byte 0xff is conflated with the value EOF = -1, so even though the input is not forbidden, it may give the wrong answer.

Casting char to unsigned int first before passing the result to ctype(3) doesn't help: inputs like -128 are unchanged by this cast, because (on a two's-complement machine with 32-bit int and unsigned int), converting the signed char with integer value -128 to unsigned int gives integer value 2^32 - 128 = 0xffffff80, which is out of range, and which is converted in int back to -128, which is also out of range.

It is necessary to cast char inputs to unsigned char first; you can then cast to unsigned int if you like but there's no need because the functions will always convert the argument to int by definition.  So the above fragment needs to be:

```
char *ptr = ...;
... isspace((unsigned char)*ptr) ...
```

When I run pkgconf on the following pkgconfig file, it crashes with SIGSEGV:

```
Name: Escape
Version: 4.2.0
Description: Escape utility library
Libs: -Llink\ path\ with\ spaces
Cflags: -Iinclude\ path\ with\ spaces -DA=\"escaped\ string\'\ literal\" -DB=ESCAPED\ IDENTIFIER -DFOX=🦊
```

gdb shows:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000703e17c0bb80 in pkgconf_parser_parse (f=f@entry=0x703e179ca040,
    data=data@entry=0x703e17e21000, ops=ops@entry=0x703e17e0fae0,
[...]

PR #292, opened 2023-03-17
Avoid undefined behaviour with the ctype(3) functions.

riastradh:
fix https://github.com/pkgconf/pkgconf/issues/291

As defined in the C standard:

        In all cases the argument is an int, the value of which shall
        be representable as an unsigned char or shall equal the value
        of the macro EOF.  If the argument has any other value, the
        behavior is undefined.

This is because they're designed to work with the int values returned by getc or fgetc; they need extra work to handle a char value.

If EOF is -1 (as it almost always is), with 8-bit bytes, the allowed inputs to the ctype(3) functions are:

        {-1, 0, 1, 2, 3, ..., 255}.

However, on platforms where char is signed, such as x86 with the usual ABI, code like

        char *ptr = ...;
        ... isspace(*ptr) ...

may pass in values in the range:

        {-128, -127, -126, ..., -2, -1, 0, 1, ..., 127}.

This has two problems:

1. Inputs in the set {-128, -127, -126, ..., -2} are forbidden.

2. The non-EOF byte 0xff is conflated with the value EOF = -1, so even though the input is not forbidden, it may give the wrong answer.

Casting char to unsigned int first before passing the result to ctype(3) doesn't help: inputs like -128 are unchanged by this cast, because (on a two's-complement machine with 32-bit int and unsigned int), converting the signed char with integer value -128 to unsigned int gives integer value 2^32 - 128 = 0xffffff80, which is out of range, and which is converted in int back to -128, which is also out of range.

It is necessary to cast char inputs to unsigned char first; you can then cast to unsigned int if you like but there's no need because the functions will always convert the argument to int by definition.  So the above fragment needs to be:

        char *ptr = ...;
        ... isspace((unsigned char)*ptr) ...

This patch changes unsigned int casts to unsigned char casts, and adds unsigned char casts where they are missing.

### DIFF
--- a/cli/bomtool/main.c
+++ b/cli/bomtool/main.c
@@ -309,7 +309,7 @@ main(int argc, char *argv[])
 		if (maximum_package_count > 0 && pkgq.length > maximum_package_count)
 			break;
 
-		while (isspace((unsigned int)package[0]))
+		while (isspace((unsigned char)package[0]))
 			package++;
 
 		/* skip empty packages */
--- a/cli/main.c
+++ b/cli/main.c
@@ -407,7 +407,8 @@ apply_env(pkgconf_client_t *client, pkgconf_pkg_t *world, void *env_prefix_p, in
 	char workbuf[PKGCONF_ITEM_SIZE];
 
 	for (it = want_env_prefix; *it != '\0'; it++)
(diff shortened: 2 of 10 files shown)


kaniini:
Applied, thanks!

Issue #298, opened 2023-06-13
The configure script doesn't exist


Erotemic:
The docs state that to build pkgconf you can use the standard unix pattern:

```
$ ./configure
$ make
$ sudo make install
```

However, there is no configure script in the main repo. There is an autoconfig based configure.ac, but that's not the same thing. I'm attempted to install pkg-config from source in a very restricted environment (I can't install system libs), and I'm having trouble figuring out how to do it.

Erotemic:
Thanks, I made an MR to help others who have the same confusion in the future. I'm not sure if you run autogen.sh with any arguments, but the docs should probably mention it as a step for someone unfamiliar with autoconf-style projects.

dg0yt:
> But it is however not noted in the README that you need to run `autogen.sh` before running `./configure.sh`.

You are meant to use the tarballs with pregenerated scripts, and then you wouldn't need to run `autogen.sh`.

PR #299, opened 2023-06-13
Add autogen step to the README

Erotemic:
Addresses my confusion in https://github.com/pkgconf/pkgconf/issues/298


kaniini:
As noted in the README, tarballs with pregenerated autoconf scripts are available at https://distfiles.dereferenced.org/pkgconf/

Issue #306, opened 2023-07-24
Output order different than what is in tests


bruchar1:
I compiled pkgconf on Windows. I tried to reproduce the test cases in `basic.sh`. Some cases give me a different output order than in the test. For instance:

```
$ build\pkgconf.exe --with-path=\env\pkgconf\tests\lib1\ --libs intermediary-1 intermediary-2                                                                                                                                  
-lintermediary-2 -lintermediary-1 -lfoo -lbar -lbaz
```
while the test says the output should be : `-lintermediary-1 -lintermediary-2 -lfoo -lbar -lbaz\n`

Do you have any idea why this order differs? Does it matter?

kaniini:
Probably fixed by 5d0e2f50923aff3ce4778bc3ab39452fc2daec5f, it turns out that qsort is not stable on Windows.

bruchar1:
I confirm it fixes the issue. Thank you!

PR #308, opened 2023-07-27
Add python bindings to libpkgconf

bruchar1:
I know there is a pkgconf-py project, but it seems outdated and unmaintained.

This is an attempt to provide Python bindings, through the cffi lib.

- low level bindings are provided for every exported function in libpkgconf.h.
- a high-level functions replicates the behavior of the cli.

I replicated the test suite in Python. However, I have issues with lib order, but it is probably related to the Windows platform (see #306), I marked temporarily those tests as xfail (but skip would probably be more appropriate in this context).

The goal of this bindings is to allow meson to call pkgconf via a Python lib rather that via subprocesses. This will dramatically improve performances, especially on Windows.


meson.build
```diff
@@ -128,6 +129,15 @@ kyuafile = configure_file(input : 'Kyuafile.in', output : 'Kyuafile', configurat
 test('kyua', kyua_exe, args : ['--config=none', 'test', '--kyuafile', kyuafile, '--build-root', meson.current_build_dir()])
 subdir('tests')
 
+
+with_python = get_option('python') \
+  .require(libtype == 'static', error_message: 'python bindings require static linking') \
+  .require(get_option('debug') == false, error_message: 'python bindings require non debug build type')
```
nekopsykose, line 135: this doesn't sound correct? why would the python bindings need a static libpkgconf? this sounds worse than just using the normal libpkgconf.so that is emitted (or allowing either), and i don't see an actual technical reason for this to be required (aside from actually writing the code to do it)

(the same goes for debug being forbidden, which really only adds -g and friends, no?)

bruchar1: For debug, it is because on Windows it would link on python3d.dll (note the d suffix), which is not installed by default. Maybe this isn't a real breaker...

For static, it is because I wasn't able to copy the dll at the right place to be installed correctly by the wheel. Maybe I did something wrong, or maybe it is specific to WIndows...

python/pypkgconf/callbacks.py
```diff
@@ -0,0 +1,105 @@
+from ._libpkgconf import ffi, lib
+
+from .constants import Flags
+from .logger import logger
+
+
+@ffi.def_extern()
+def error_handler(msg: str, client, data) -> bool:
```
tristan957, line 8: It would be a great idea to add types to to everything

bruchar1: yes. It is just not super-obvious with the cffi bindings. But I could try... But on another hand, I don't want to spend too much time on this PR if it is not to be adopted.

meson.build
nekopsykose: hmm.. generally the wheels don't necessarily contain the .so. i don't know what one gets from pypi, but on distros cffi would load a regular system library in everything i've seen. it would be 'weird' to require the static linkage. so, that does all sound windows specific, for which to best handle you'd want to write it to work with either, so static can be used on windows, and linux people can pick whatever.

for python3d.dll, that is indeed a windows thing. on linux there is no such thing- the debug only implies that you add some debugging flags and whatnot and not that a different python library is linked (which is wholly separate and may or may not have debug, might have them split symbols elsewhere, ..., and be out of control of the build system here at all). so that is only valid when host=windows (or at least not for host=linux).

python/pypkgconf/callbacks.py
tristan957: Sure, let's wait and see what the maintainer says.

meson.build
bruchar1: I reworked the meson files a little bit. However, it would require a feature not yet in meson for it to work with shared library on Windows. Other cases should be ok (but not tested...)

kaniini:
I do not think pkgconf should include bindings for other languages.  These should be distributed as separate projects.

Issue #310, opened 2023-08-04
--modversion should accept multiple arguments when it's the same package


ptrcnull:
currently, the following command will fail on pkgconf 2.0:
```
$ pkg-config --modversion dbus-1 'dbus-1 >= 1.6'
pkgconf: --modversion requested with multiple packages, output would be ambiguous
```

while it might be a bit redundant to do so in this case, the output would not be ambiguous, and unfortunately there's [prior art][1]

furthermore, with the check in place, there doesn't seem to be a way to query a package with both an upper and lower version bound; it didn't really work on pkgconf 1.x either, but it seems to work with pkg-config, so maybe it's worth looking into as well:
```
root@86a8a87bcfe7:/# pkg-config --modversion 'dbus-1 >= 1.6'
1.12.16
root@86a8a87bcfe7:/# pkg-config --modversion 'dbus-1 >= 1.6' 'dbus-1 < 1.10'
Requested 'dbus-1 < 1.10' but version of dbus is 1.12.16
root@86a8a87bcfe7:/# pkg-config --modversion 'dbus-1 >= 1.6' 'dbus-1 < 1.15'
1.12.16
1.12.16
```

[1]: https://github.com/rust-lang/pkg-config-rs/issues/150

nirbheek:
This seems like completely unnecessary breakage. According to the release notes, this is an error because it's ambiguous. But the output is not ambiguous at all. The versions are printed in the order of the arguments:

```
$ pkg-config --modversion sqlite3 zlib
3.39.5
1.2.11
```

IMO this change should be completely reverted. There is no good reason for diverge from pkg-config in this case.

Polynomial-C:
Unfortunately with commit a97b75ab2c1d031982c35a4886102413e4ec8eee being reverted, `pkgconf-2.0.0` only returns one version:
```
> pkgconf --modversion sqlite3 zlib
3.42.0
```
So there's some more breakage compared to pkgconfig.

nekopsykose:
> Unfortunately with commit https://github.com/pkgconf/pkgconf/commit/a97b75ab2c1d031982c35a4886102413e4ec8eee being reverted, pkgconf-2.0.0 only returns one version:  

that was always like that, 1.9.5 behaves the same. 2.0 with that reverted is a safe upgrade

Polynomial-C:
[...]

PR #311, opened 2023-08-16
cli: fix --modversion not working with provided deps

breavyn:
If a dep is resolved via the provides property the package id will be different to the requested dep. This matches the requested dep against provided dep names instead.

I ran into this issue when building a package depending on sdl, which is provided by sdl12_compat on my system. I'm not familiar enough with pkgconf to say this change couldn't have other unintended side effects, but it has worked for me in this case.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -288,6 +288,34 @@ apply_digraph(pkgconf_client_t *client, pkgconf_pkg_t *world, void *unused, int 
 }
 #endif
 
+static void
+print_modversion(pkgconf_pkg_t *world, pkgconf_queue_t *queue_node)
+{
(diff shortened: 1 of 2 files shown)


eli-schwartz:
This is also a fix for https://github.com/pkgconf/pkgconf/issues/310#issuecomment-1677844842 ?

eli-schwartz:
Hmm, it appears not.

breavyn:
@eli-schwartz I'm not sure it's going to be possible to catch all the edge cases here without adding the requested package name to each package as its resolved.

breavyn:
Closing this in favour of #313.

Issue #312, opened 2023-08-18
pkgconf --modversion /path/to/foo.pc is broken


floppym:
Before f49029c681653044d19450f6c278bf54543d0d18:
```
% pkgconf --modversion /usr/share/pkgconfig/udev.pc
254
```
After f49029c681653044d19450f6c278bf54543d0d18:
```
% pkgconf --modversion /usr/share/pkgconfig/udev.pc
(no output)
```

eli-schwartz:
Using the PR by @breavyn at #311:
```diff
diff --git a/cli/main.c b/cli/main.c
index c967455..8f9d9cf 100644
--- a/cli/main.c
+++ b/cli/main.c
@@ -302,6 +302,7 @@ print_modversion(pkgconf_pkg_t *world, pkgconf_queue_t *queue_node)
 		{
 			pkgconf_dependency_t *provided_dep = provided_iter->data;
 
+			printf("testing id: %s, provided package: %s, node package: %s \n", pkg->id, provided_dep->package, queue_node->package);
 			if (strcmp(provided_dep->package, queue_node->package))
 				continue;
 
```

```bash
$ ./pkg/bin/pkgconf --modversion /usr/share/pkgconfig/udev.pc
testing id: udev, provided package: udev, node package: /usr/share/pkgconfig/udev.pc 
```

kaniini:
#313 has been merged

PR #313, opened 2023-08-18
cli: fix --modversion not showing version in various cases

breavyn:
There are numerous edge cases where version is wrong or missing when matching the dependency queue to resolved packages. This adds the dependency name as it appears in the dependency queue to each package as it is resolved, allowing for a simple and correct comparison when printing.

This is an alternative approach to  #311

This should resolve these current issues;

- duplicate and/or incorrect versions shown in cases like libev libevdev
- no version shown for complete paths (#312)
- no version shown when package is matched by provides. sdl  -> sdl12_compat
- no version shown when package is matched by -uninstalled


breavyn:
Added additional regression tests demonstrating these cases.

PR #315, opened 2023-09-03
libpkgconf: fix memory leak

breavyn:
Passing all tests (with asan).

This section can receive the same package instance multiple times. From what I've seen `pkgdep->package` is always the same, so I've elected to only set the `why` field the first time.


kaniini:
Applied, thanks

PR #319, opened 2023-09-07
cli: fix --modversion not showing version with versioned query

### LINKED ISSUE #317: --modversion changed in incompatible way in 2.0.3
Basically, on versions older than 2.0.3 (tested with 2.0.2 and 1.8.0), `pkg-config --modversion "somelibrary "` (note the extra space) worked exactly the same as `pkg-config --modversion somelibrary` (note lack of extra space).

As if 2,0.3, this behavior has changed. Now, `pkg-config --modversion "somelibrary "` doesn't report any version at all.

Ideally, this wouldn't be an issue. But it looks like some projects have accidentally been relying on this behavior. I understand that those projects should ideally be changed. However, this is an unexpected and undocumented change in behavior for pkgconf. So I'm bringing it to your attention :)

xry111:
Fixes #317.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -306,7 +306,7 @@ apply_modversion(pkgconf_client_t *client, pkgconf_pkg_t *world, void *data, int
 			pkgconf_dependency_t *dep = world_iter->data;
 			pkgconf_pkg_t *pkg = dep->match;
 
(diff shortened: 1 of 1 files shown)


kaniini:
Applied, thanks.

PR #323, opened 2023-09-13
meson: Add with-system options

### LINKED ISSUE #283: Building pkgconf with meson doesn't allow to override the system folders
Hi!

I'm writing to report an issue I found when using pkgconf with Meson https://github.com/mesonbuild/meson/issues/11088.

In short, I build pkgconf along with the rest of the dependencies in a root I control. But when I use it, due to the stripping behaviour, said root will get removed from the include folders, and cause dependent libraries to fail to build or link.

It would be easy to work around this with `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` (which I use locally), but would it be possible to allow the Meson build system to use the same flags as the autotools system?

(Please note that I'm on Windows, hence why it's not possible to use a POSIX shell.)

amyspark:
Hi @kaniini,

This PR is to allow the Meson build system to override the system include and/or lib folder like Autotools.

Fixes #283

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -49,8 +49,18 @@ foreach f : ['libdir', 'datadir']
   personality_path += [join_paths(get_option('prefix'), get_option(f), 'pkgconfig', 'personality.d')]
 endforeach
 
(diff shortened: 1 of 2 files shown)


kaniini:
Applied, thanks

Issue #322, opened 2023-09-11
Library order is still not correct


stanhu:
@kaniini This was supposedly fixed in pkgconf v2.0 (#268), but I'm still seeing an issue. Attached is the reproduction step with the `.pc` files I'm trying to use:

```shell
tar xvf test-pc.tar.gz
PKG_CONFIG_PATH=pkgconfig pkg-config --libs re2
```

[test-pc.tar.gz](https://github.com/pkgconf/pkgconf/files/12573655/test-pc.tar.gz)

On Windows, with `pkgconf` v2.0.3 I see the incorrect order. For example, notice that `-labsl_spinlock_wait` precedes `-labsl_flags`, but in v1.8.0 `-labsl_spinlock_wait` is towards the end as it should be:

```
-pthread -L/home/stanhu/github/re2/ports/x64-mingw-ucrt/x86_64-w64-mingw32/libre2/2023-09-01/lib -lr
e2 -L/home/stanhu/github/re2/ports/x64-mingw-ucrt/x86_64-w64-mingw32/abseil/20230802.0/lib -labsl_sp
inlock_wait -labsl_flags -labsl_bad_optional_access -labsl_string_view -labsl_strings_internal -labs
l_flags_program_name -labsl_graphcycles_internal -labsl_kernel_timeout_internal -labsl_civil_time -l
absl_time_zone -labsl_symbolize -ldbghelp -labsl_debugging_internal -labsl_demangle_internal -labsl_
malloc_internal -labsl_flags_internal -labsl_flags_marshalling -labsl_log_severity -labsl_str_format
_internal -labsl_flags_reflection -labsl_flags_private_handle_accessor -labsl_flags_commandlineflag
-labsl_flags_commandlineflag_internal -labsl_flags_config -labsl_cord -labsl_crc32c -labsl_crc_inter
nal -labsl_crc_cpu_detect -labsl_cordz_functions -labsl_cordz_handle -labsl_stacktrace -labsl_cord_i
nternal -labsl_cordz_info -labsl_crc_cord_state -labsl_city -labsl_bad_variant_access -labsl_low_lev
el_hash -labsl_int128 -labsl_hash -labsl_hashtablez_sampler -labsl_exponential_biased -labsl_time -l
absl_raw_logging_internal -labsl_throw_delegate -labsl_base -ladvapi32 -labsl_raw_hash_set -labsl_st
rings -labsl_synchronization
```

However, with `pkgconf` v1.8.0, I see the correct order:

[...]

PR #327, opened 2023-10-03
Fix search order

### LINKED ISSUE #326: Issue with PKG_CONFIG_PATH resolving past 2.0.0
Steps to reproduce:
1. Prep two hypothetical prefixes (could be a local install and the normal `/usr/lib/pkgconfig` folder)
```bash
for folder in a b; do
    mkdir -p $folder
    sed 's/PREFIX/'$folder'/' > $folder/c.pc << 'EOF'
prefix=/PREFIX
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: c
Description: c
Version: 1.0.0
Cflags: -I${includedir}
Libs: -L${libdir} -lc
Libs.private: -ld
EOF
done
export PKG_CONFIG_PATH="$PWD/a:$PWD/b"
```

2. run pkgconf < 2.0.0
```bash
$ pkgconf --static --libs c
-L/a/lib -lc -ld
```
3. run pkgconf >= 2.0.0
```bash
$ ./bui/pkgconf --static --libs c
-L/b/lib -lc -ld
```

Offending commit after bisecting seems to be https://github.com/pkgconf/pkgconf/commit/384ade5f316367f15da9dbf09853e06fe29bf2bf and reverting that restores pre-2.0.0 behavior, however, I am not entirely sure the motivation behind that commit, so I can't tell if reverting will break something else.

I can't submit to <https://gitea.treehouse.systems/ariadne/pkgconf> right now, as it's giving me error 500.

dg0yt:
Revert https://github.com/pkgconf/pkgconf/commit/384ade5f316367f15da9dbf09853e06fe29bf2bf. (Credits: @1480c1.)
Always insert `--with-path` dirs (command line) before anything else (environment, defaults).
Add tests.

Fixes #326.

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -1019,7 +1019,7 @@ main(int argc, char *argv[])
 	pkgconf_client_init(&pkg_client, error_handler, NULL, personality);
 
 	/* user-supplied dirs come first */
-	pkgconf_client_dir_list_add(&pkg_client, &prepended_dir_list);
+	pkgconf_path_copy_list(&pkg_client.dir_list, &prepended_dir_list);
 	pkgconf_path_free(&prepended_dir_list);
 
 #ifndef PKGCONF_LITE
--- a/cli/main.c
+++ b/cli/main.c
@@ -827,7 +827,7 @@ main(int argc, char *argv[])
 {
 	int ret;
 	pkgconf_list_t pkgq = PKGCONF_LIST_INITIALIZER;
-	pkgconf_list_t dir_list = PKGCONF_LIST_INITIALIZER;
+	pkgconf_list_t prependend_dir_list = PKGCONF_LIST_INITIALIZER;
 	char *builddir;
 	char *sysroot_dir;
 	char *env_traverse_depth;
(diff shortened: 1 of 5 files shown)


lazka:
Just wondering, why add new public APIs with this? This seems unrelated to the issue.

dg0yt:
Do you propose to drop `pkgconf_client_dir_list_add` and to access `client->dir_list` directly from `main.c`?
(I guess my C++ish bias lets me avoid direct access to members.)

dg0yt:
> to drop `pkgconf_client_dir_list_add` and to access `client->dir_list` directly from `main.c`?

Done.

chewi:
This sadly didn't help in the [libxml2 case](https://gitlab.gnome.org/GNOME/libxml2/-/issues/602).

vishwin:
I can confirm that reverting 384ade5 fixes what I [reported on gitea](https://gitea.treehouse.systems/ariadne/pkgconf/issues/250), which is most likely the same issue(s) reported here.

dg0yt:
> This sadly didn't help in the [libxml2 case](https://gitlab.gnome.org/GNOME/libxml2/-/issues/602).

Understood. Remove the link to #322.

kaniini:
Fixed by 652aff97906d7eadca6fe727b16f6511d8dbfe05

PR #330, opened 2023-10-31
Search XDG Base Directories for personality files

### LINKED ISSUE #329: Search personality files in XDG_DATA_DIRS
Depending on how pkgconf is built, personalities are looked for (with normal prefixes) in /usr/share/pkgconfig/personality.d and /etc/pkgconfig/personality.d (https://github.com/pkgconf/pkgconf/blob/752a9825dc8660d247c457aa4d256db4800ebd7c/configure.ac#L34-L37) or /usr/lib/pkgconfig/personality.d and /usr/share/pkgconfig/personality.d (https://github.com/pkgconf/pkgconf/blob/752a9825dc8660d247c457aa4d256db4800ebd7c/meson.build#L47-L50). 

I had hoped that symlinking pkgconf into a toolchain directory's bin subdirectory as `arm-linux-musleabihf-pkg-config` and setting `XDG_DATA_DIRS` would make pkgconf search for personality files (also) in the `pkgconfig/personality.d` subdirectory of every entry in `XDG_DATA_DIRS`. Would this be something you'd be willing to add and/or accept a PR for?

`XDG_DATA_DIRS` is specfied in the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

viccie30:
Follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) when searching for personality files.

Would fix #329.

### DIFF
--- a/libpkgconf/personality.c
+++ b/libpkgconf/personality.c
@@ -241,17 +241,19 @@ personality_warn_func(void *p, const char *fmt, ...)
 }
 
 static pkgconf_cross_personality_t *
-load_personality_with_path(const char *path, const char *triplet)
+load_personality_with_path(const char *path, const char *triplet, bool datadir)
 {
 	char pathbuf[PKGCONF_ITEM_SIZE];
 	FILE *f;
 	pkgconf_cross_personality_t *p;
 
 	/* if triplet is null, assume that path is a direct path to the personality file */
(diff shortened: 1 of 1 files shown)


kaniini:
applied, thanks

Issue #343, opened 2024-02-06
incorrect library order with pkgconf 2.1.1


dg0yt:
With pgkconf 2.1.1, link libraries are emitted in wrong order for `pangocairo` in vcpkg. `libpcre2-8` should be emitted after `glib-2.0` but isn't. However, in the digraph output, libpcre2-8 is listed "after" glib-2.0 in both versions. Digraphs are effectively identical for both versions of pkgconf, only the order of edges is different.

2.1.0:
![img-2 1 0](https://github.com/pkgconf/pkgconf/assets/13567791/b00266f8-a257-4080-8ea4-e3ebcafd01a5)

2.1.1:
![img-2 1 1](https://github.com/pkgconf/pkgconf/assets/13567791/2762e050-7eda-489b-a4da-63bb655f0696)

...

dg0yt:
Hm, can't attach a zip or tar.gz now...

dg0yt:
Certainly broken by #339.
Now that nodes are not visited more than once, the effect is that a node's fragments are emitted emitted only once. Fragments unique to a single pc file (like `-lpcre2-8` in `libpcre2-8.pc`) *never* become subject to merge-back - but *merge-back is required* to reflect all dependencies.
(Merge-back still happens for *common* fragments like `-lm`.)

thesamesam:
cc @qazwsxedcrfvtg14

PR #344, opened 2024-02-09
Fix lib order regression

dg0yt:


kaniini:
I disagree that this fixes #343, as it only fixes it by removing a needed optimization.  Instead we should try to figure out how to apply the optimization in such a way that generates the desired order of edges in the dependency graph.

dg0yt:
Maybe it doesn't fix #343, but is not an "optimization" if degrades results from "slow" to "wrong". And it wouldn't have passed the tests which I added now.
I'm already studying improvements, but all I have is the code. I understood that the "optimization" added the third serial number, after `identifier` and `serial`. The flattening operates on `idenifier`. This is not enough as long as it represents first-visit order, due to the "optimization".

kaniini:
Right, sorry I should rephrase.  We should try to fix the underlying performance goal while ensuring that the dependency graph is evaluated correctly.

dg0yt:
Some conclusion from experiments:
- The expensive part is probably not the walk over all paths, but the unlimited copying of dependencies into virtual:world every time a node is visited.
- To determine a valid serial order, a full walk over all paths is still the best/simplest approach.
- IMO the solution shouldn't be split into a "regular" and a "private" part. It is *one* graph once `.private` is included in traversal, so it is only *one* serialization. IOW `.private` is only for input, not for output.
  (What might be tracked for later filtering is whether a package is reachable without private deps or not.)
  However, this would change some outputs (also for tests).

kaniini:
The first is testing undefined behavior (an unsolvable dependency cycle) and is only meant to see if the solver behavior has changed.  We expect a change so it’s fine here.

The rest appear functionally equivalent and arguably better.

dg0yt:
Cf. #344

PR #339, opened 2023-12-22
Use traverse_id to prevent traverse pkgdep twice

qazwsxedcrfvtg14:
Context: https://issuetracker.google.com/issues/317357322

In the ChromeOS build system, the pkg-config is very slow because we have a very complex dependency graph.

After deep diving into the code of pkg-config, I think it would not work very well when we have a "deep" dependency graph.

For example:
```
O
|\
| \
O  O
|\/|
|/\|
O  O
|\/|
|/\|
O  O
...
|\/|
|/\|
O  O
|\/|
|/\|
O  O
```

If we have this kind of dependency graph, the time complexity of pkg-config will become `O(2^N)`, that is not an acceptable time complexity.

To fix the issue, I add a traverse_id into the client and root to make sure we don't traverse a package multiple times.

PR #345, opened 2024-02-15
Revise traversal and flattening

dg0yt:
Fixes lib order regressions.  
Limit (allocationwise) expensive traversal steps while still doing necessary steps for determining the topological sort.

**Major change:**
The solution is put as a single list into `requires` - also when walking over private dependencies.  
In this list, dependencies which result only from private paths are marked as such.
(NB: A walk with private dependencies only adds restrictions and removes possible solutions. So skipping the private dependencies of a complete solution yields a valid solution for the problem without private dependencies.)

### DIFF
--- a/libpkgconf/libpkgconf.h
+++ b/libpkgconf/libpkgconf.h
@@ -137,8 +137,7 @@ struct pkgconf_path_ {
 #define PKGCONF_PKG_PROPF_CACHED		0x02
 #define PKGCONF_PKG_PROPF_UNINSTALLED		0x08
 #define PKGCONF_PKG_PROPF_VIRTUAL		0x10
-#define PKGCONF_PKG_PROPF_VISITED		0x20
-#define PKGCONF_PKG_PROPF_VISITED_PRIVATE	0x40
+#define PKGCONF_PKG_PROPF_PRIVATE		0x40
 
 struct pkgconf_pkg_ {
 	int refcount;
--- a/libpkgconf/libpkgconf.h
+++ b/libpkgconf/libpkgconf.h
@@ -176,7 +176,6 @@ struct pkgconf_pkg_ {
 	pkgconf_tuple_t *orig_prefix;
 	pkgconf_tuple_t *prefix;
 
-	uint64_t serial;
 	uint64_t identifier;
 	uint64_t traverse_serial;
 };
@@ -213,8 +212,8 @@ struct pkgconf_client_ {
 
 	bool already_sent_notice;
 
+	/* a counter to create identifiers for pkg nodes and to identify traversals */
 	uint64_t serial;
-	uint64_t identifier;
 	uint64_t traverse_serial;
 
 	pkgconf_pkg_t **cache_table;
--- a/libpkgconf/queue.c
+++ b/libpkgconf/queue.c
@@ -258,12 +258,12 @@ pkgconf_queue_verify(pkgconf_client_t *client, pkgconf_pkg_t *world, pkgconf_lis
 	 */
 	++client->serial;
 
-	PKGCONF_TRACE(client, "flattening requires deps");
+	PKGCONF_TRACE(client, "flattening 'Requires' deps");
 	flatten_dependency_set(client, &world->required);
 
 	++client->serial;
 
-	PKGCONF_TRACE(client, "flattening requires.private deps");
+	PKGCONF_TRACE(client, "flattening 'Requires.private' deps");
 	flatten_dependency_set(client, &world->requires_private);
 
 	return PKGCONF_PKG_ERRF_OK;
--- a/libpkgconf/queue.c
+++ b/libpkgconf/queue.c
@@ -162,7 +162,7 @@ flatten_dependency_set(pkgconf_client_t *client, pkgconf_list_t *list)
 		if (pkg == NULL)
 			continue;
 
-		if (pkg->serial == client->serial)
+		if (pkg->identifier == client->serial)
 		{
 			pkgconf_node_delete(node, list);
 			pkgconf_dependency_unref(client, dep);
@@ -189,7 +189,7 @@ flatten_dependency_set(pkgconf_client_t *client, pkgconf_list_t *list)
 			}
 		}
 
-		pkg->serial = client->serial;
+		pkg->identifier = client->serial;
 
 		/* copy to the deps table */
 		dep_count++;
--- a/libpkgconf/queue.c
+++ b/libpkgconf/queue.c
@@ -163,7 +163,7 @@ flatten_dependency_set(pkgconf_client_t *client, pkgconf_list_t *list)
 {
 	pkgconf_node_t *node, *next;
 	pkgconf_dependency_t **deps = NULL;
-	size_t dep_count = 0, i;
+	size_t dep_count = 0, i, dep_allocation = 0;
 
 	PKGCONF_FOREACH_LIST_ENTRY_SAFE(list->head, next, node)
 	{
@@ -204,7 +204,11 @@ flatten_dependency_set(pkgconf_client_t *client, pkgconf_list_t *list)
 
 		/* copy to the deps table */
 		dep_count++;
-		deps = pkgconf_reallocarray(deps, dep_count, sizeof (void *));
+		if (dep_count > dep_allocation)
+		{
+			dep_allocation = dep_count + 31;
+			deps = pkgconf_reallocarray(deps, dep_allocation, sizeof (void *));
+		}
 		deps[dep_count - 1] = dep;
 
 		PKGCONF_TRACE(client, "added %s to dep table", dep->package);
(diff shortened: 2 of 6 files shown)


dg0yt:
~~~
PKG_CONFIG_PATH=tests/lib1 build/pkgconf --static baz --digraph | dot -Tpng -obaz.png
~~~
`baz.pc`: `foo` is a private dep, result:
baz

~~~
PKG_CONFIG_PATH=tests/lib1 build/pkgconf --static foo baz --digraph | dot -Tpng -ofoo-baz.png
~~~
`foo` is now public (on the command line), result:
foo-baz

libpkgconf/libpkgconf.h
```diff
@@ -137,8 +137,7 @@ struct pkgconf_path_ {
 #define PKGCONF_PKG_PROPF_CACHED		0x02
 #define PKGCONF_PKG_PROPF_UNINSTALLED		0x08
 #define PKGCONF_PKG_PROPF_VIRTUAL		0x10
-#define PKGCONF_PKG_PROPF_VISITED		0x20
-#define PKGCONF_PKG_PROPF_VISITED_PRIVATE	0x40
+#define PKGCONF_PKG_PROPF_PRIVATE		0x40
```
dg0yt, line 140: Hm, should this be `0x20` instead?

libpkgconf/queue.c
```diff
@@ -249,23 +264,24 @@ pkgconf_queue_verify(pkgconf_client_t *client, pkgconf_pkg_t *world, pkgconf_lis
 		return result;
 	}
 
+	/* determine identifiers for topological sort */
```
qazwsxedcrfvtg14, line 267: IIUC, this is not the topological order, this is the node visited order.
The topological order is the "reversed node exited" order.

qazwsxedcrfvtg14: What I'm thinking about is something like this:
https://github.com/qazwsxedcrfvtg14/pkgconf/commit/85daab772a356597570e076aead44adfdbacbedc

qazwsxedcrfvtg14:
I use this script to create 200 .pc files under the `tests` folder:

```bash
#!/bin/bash
size=100
for k in {0..1}; do 
  for i in $(seq 0 $size); do
    echo "Name: chain${k}-${i}
    Description: test
    Version: 1.0
    Requires: chain0-$((i+1)) chain1-$((i+1))
    Libs: -lchain${k}_${i}
    Libs.private: -lchain${k}_${i}_private" > lib4/chain${k}-${i}.pc
  done
  echo "Name: chain${k}-$((size+1))
  Description: test
  Version: 1.0
  Libs: -lchain${k}_$((size+1))
  Libs.private: -lchain${k}_$((size+1))_private" > lib4/chain${k}-$((size+1)).pc
done
```

And this command will run forever with this PR:
```
PKG_CONFIG_PATH=./lib4 ../pkgconf --static --libs-only-l chain0-0 chain1-0
```

dg0yt:
Thanks for the script. A few days more of waiting, and I would have written something similar to model the pattern from https://github.com/pkgconf/pkgconf/pull/339.

> IIUC, this is not the topological order, this is the node visited order.
> The topological order is the "reversed node exited" order.

🤔 Hm... Hm... Okay, I start to see my fault.
The key for O(V+E) is to update the identifier on "node exit", i.e. after walking the child lists.
This is not what my changes do ATM.

> What I'm thinking about is something like this: [qazwsxedcrfvtg14@85daab7](https://github.com/qazwsxedcrfvtg14/pkgconf/commit/85daab772a356597570e076aead44adfdbacbedc)

I can't say much from the GH display, but it seems to point into the right direction, "node exit" for order.
Analysing public reachability may need a different grip now, but it can't be expensive.

dg0yt:
Yes, the node-exit modifications lead to proper runtime. 

I was about to write that it leads to changed output but formally correct results. If not ordered by dependencies, the fragments for the first exited leaf are printed last, etc. Potentially (too) surprising to humans which expect some kind of a "stable sort", in particular when trying to look for link errors.
However, while reviewing the now-failing tests, I realize that this change actually causes wrong results in combination with fragment merge-back.

https://github.com/pkgconf/pkgconf/blob/b6e04e2d47b4ca093632b9efde6303a137ea0634/tests/parser.sh#L157-L163
https://github.com/pkgconf/pkgconf/blob/b6e04e2d47b4ca093632b9efde6303a137ea0634/tests/parser.sh#L173-L179
The trailing `-lfoo` gets lost because merge-back expects stable sort, too.
If I put `foo` before `flag-order-3`, I get the original result.

So node-exit ordering is right, but for stable ordering and merge-back, we need to visit each nodes deps in reverse order.

Given that
- there are nodes (packages) and edges (dependencies),
- order is partially encoded in lists of edges,
- each node is visited once,
- the desired topological sort is the node exit order (post-order)
- `pkgconf_pkg_traverse` calls a callback function on node entry (pre-order)
- the callback named `pkgconf_queue_collect_dependents` actually
  - collects dependencies (i.e. copies edges pointing to independent nodes),
  - creates multiple edges for the same node (package), requiring filtering,
  - being used pre-order, creates a list with wrong node order, requiring sorting,
- the way of creating serials puts early leafs to the end of the result (which is valid from format POV but 

... I ~wonder if there isn't~ realize that there is already too much mismatch between goals and tools, and ~if the solution shouldn't~ that the solution should be constructed using a different traversal than the one used for output purposes.

dg0yt:
FTR: Modified generator script with `Requires.private`:
~~~
#!/bin/sh
size=100
mkdir -p lib4 &&
for k in 0 1; do 
  for i in $(seq 0 $size); do
    echo "Name: chain${k}-${i}
    Description: test
    Version: 1.0
    Requires: chain0-$((i+1))
    Requires.private: chain1-$((i+1))
    Libs: -lchain${k}_${i}
    Libs.private: -lchain${k}_${i}_private" > lib4/chain${k}-${i}.pc
  done
  echo "Name: chain${k}-$((size+1))
  Description: test
  Version: 1.0
  Libs: -lchain${k}_$((size+1))
  Libs.private: -lchain${k}_$((size+1))_private" > lib4/chain${k}-$((size+1)).pc
done
~~~

Output for `pkgconf --with-path=lib4 chain0-95 --static --digraph | dot -Tpng -oimg
.png`:
img

On my PC, runtime gets interesting between 85 and 80.

dg0yt:
> the solution should be constructed using a different traversal than the one used for output purposes.

Done. No problem left with `pkgconf --with-path=lib4 chain0-0 --static --digraph`.  
`libpkgconfig.h` is more similar to 2.1.0.

[1cb22dd](https://github.com/pkgconf/pkgconf/pull/345/commits/1cb22dd59fea23700ae14ceea980ac7ca313655a) is the key change.
[39021f0](https://github.com/pkgconf/pkgconf/pull/345/commits/39021f0dcdc6017cd7d8c2101e02b0d6dede19bf) is a bonus, for  `pkgconf --libs tests/lib1/circular-3.pc`solution digraph changing from  
circular-before to circular-after
[e051f51](https://github.com/pkgconf/pkgconf/pull/345/commits/e051f5193818dd88a9ce22c99dbda38d7a2f9219) is bonus.

dg0yt:
Successfully tested in vcpkg CI again ([x64-linux](https://dev.azure.com/vcpkg/public/_build/results?buildId=99825&view=results), [x64-windows](https://dev.azure.com/vcpkg/public/_build/results?buildId=99806&view=results)).

kaniini:
This is incredible work, and is aligned with my original thinking to solve this problem, but I didn’t have the time to go through and do it.

Since it involves changes to the solver, we should probably cut pkgconf 2.2.0 with these changes once everyone has confidence in them.

What do you think, @qazwsxedcrfvtg14?  Do these changes work for you?

dg0yt:
Thanks. I really think I finally understood the plan :-) 

It is probably advisable to raise at least the minor version, not only the patch version, given the ins and outs in flags/enums. If the lib is a thing for external users.

dg0yt:
> It is probably advisable to raise at least the minor version, not only the patch version, given the ins and outs in flags/enums. If the lib is a thing for external users.

So the lib is a thing, and so is the ABI: #347.

Maybe it there should even be a release candidate.

qazwsxedcrfvtg14:
The changes looks good to me!

dg0yt:
All this beautiful digraph pictures. But they had so many edges from the root virtual:world, and didn't really show the order of the packages in the output. So I just tried a change to `--digraph` to show the order of packages in the result as red edges. I didn't expect that it shows the effect of this PR so dramatically, given a suitable complex input such as (static) pango.

Before this PR ... let's call it a piece of art.
img

With this PR ... a clear order.
img3

dg0yt:
@kaniini Can this be moved forward? 
AFAICS we still can only choose between 2.x releases with incorrect results or with pkgconf exploding over abseil.

kaniini:
@dg0yt yeah, I am in the process of moving some of my equipment this week, but I will cut a release next week!

dg0yt:
This PR should also resolve https://gitea.treehouse.systems/ariadne/pkgconf/issues/229.

dg0yt:
Thanks a lot for the maintainer job @kaniini. Hope this release makes everbody happy again...

PR #348, opened 2024-03-16
Mark user input in error message

dg0yt:
Before:
~~~
Package 'zlib', required by 'virtual:world', not found
~~~
After:
~~~
Package 'zlib' not found
~~~

Edit: Omitting `required by ` for virtual nodes, approximating https://github.com/pkgconf/pkgconf/pull/348#issuecomment-2012109116.


kaniini:
I wonder if it would be better to just change the message entirely if the parent is the root node?

AFAIK, original pkg-config just elides the “requested by” part for requested dependencies.

dg0yt:
Okay, there is no hurry with this ui issue.

PR #349, opened 2024-03-18
Print digraph for user request perspective

dg0yt:
Instead of generating an image with edges from the root  "virtual world" to all other nodes without carrying the resulting order, start from the list requested by the user and show final order via red eges.

Output for private-libs-duplication test case:

| before | after | 
|---------|---------|
| test-before | test-after |

We might want to change the edge colors or style to make the topological-sort edges less prominent, e.g.: red for private dependency, gray for topological sort.

Not that there is no direct visualization of the nodes are only reached via private dependencies. (This used to be encoded in unique gray edges from virtual:world to node.)

---

This PR is an extended version of the generator for https://github.com/pkgconf/pkgconf/pull/345#issuecomment-2002618450 which was still based on virtual world.
With #345 and this PR, a graph for (vcpkg) `pkgconf --static --digraph zlib pango harfbuzz` looks like this:

img4

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -303,7 +303,10 @@ apply_digraph(pkgconf_client_t *client, pkgconf_pkg_t *world, void *data, int ma
 	PKGCONF_FOREACH_LIST_ENTRY(list->head, iter)
 	{
 		pkgconf_queue_t *pkgq = iter->data;
-		printf("\"user:request\" -> \"%s\" [fontname=Sans fontsize=8]\n", pkgq->package);
+		pkgconf_pkg_t *pkg = pkgconf_pkg_find(client, pkgq->package);
+		printf("\"user:request\" -> \"%s\" [fontname=Sans fontsize=8]\n", pkg == NULL ? pkgq->package : pkg->id);
+		if (pkg != NULL)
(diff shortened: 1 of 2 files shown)


kaniini:
This is really nice work!

dg0yt:
Yes, it helps me a lot ;-) It still needs some tuning, I can't finish it before #345.

dg0yt:
My todos:

Must-have:
- Deal with all types of user input, including direct pc filepaths.

Nice to have:
- Mark private-only nodes, e.g. gray fill. (I played with putting them in a box, but this doesn't always lead to a nice layout.)
- Review edge colors/style.

dg0yt:
Maybe it is not so bad with just gray for private-only nodes.
~~~
pkgconf --static --digraph /usr/lib/x86_64-linux-gnu/pkgconfig/gnutls.pc libtasn1 | dot -Tpng -oimg.png
~~~

| libgnutls, only this PR | libgnutls, #345 + this PR | 
|--|--|
| gnutls-preliminary | gnutls-after-gh-345 |

kaniini:
Merged!

Issue #350, opened 2024-04-01
Widespread issues with empty PKG_CONFIG


ismaell:
Too many projects' `configure` scripts don't check for empty `PKG_CONFIG`, yet use the variable directly, resulting in weird error messages through the run, which often confuse users. It would be useful to either error-out or set `PKG_CONFIG=false`, so that calls result in a fail without bogus errors.

In order to support both scenarios, I propose the `PKG_PROG_PKG_CONFIG` macro:
* by default abort the configure script.
* take an extra (optional) argument to specify a custom action instead.

eli-schwartz:
I think that this is a terrible proposal and the linked commit should be reverted and re-released.

It causes an immediate regression for the *common* use case of `pkg.m4`, which is to not use PKG_PROG_PKG_CONFIG at all, but use:

```m4
PKG_CHECK_MODULES([FOO], [foo], , [
    dnl fall back to probing the library itself
    AC_CHECK_HEADER([foo.h])
    AC_CHECK_LIB([foo]) 
])
```

All these existing, ***correctly written*** autoconf scripts, now become *invalid* as a result of this change. Making existing correctly-written projects break, is much much worse than permitting broken projects to remain broken.

Broken projects should simply fix their brokenness. In fact they aren't even "broken" at all, since "weird error messages through the run, which often confuse users" is still an error, and users are empowered to report a bug to the project in question.

thesamesam:
Reverted in https://github.com/pkgconf/pkgconf/commit/ad0d8b3944cba18f8dba2e33780e4196e1305934.

@kaniini is there a chance of a point release with this in? We can backport it but we need it fixed in dist tarballs too unfortunately.

PR #351, opened 2024-04-01
pkg.m4: Abort by default if no pkgconf

ismaell:
Provide a second optional argument to PKG_PROG_PKG_CONFIG with an action to take if no pkgconf is found, and default to aborting.

Too many configure scripts in the wild don't handle an empty PKG_CONFIG, resulting in weird error messages through the run, which often confuse users.

Authors wishing to fallback to other methods can either return to the old behavior by specifying [:] as an action, or use [PKG_CONFIG=false] and call pkg-config macros normally, handling the fallback as any other failure.

Users can override PKG_CONFIG in the command line so this should imply no regression.

Closes: https://github.com/pkgconf/pkgconf/issues/350
Ref: https://gitlab.freedesktop.org/pkg-config/pkg-config/-/issues/69

### DIFF
--- a/pkg.m4
+++ b/pkg.m4
@@ -46,8 +46,8 @@ m4_if(m4_version_compare(PKG_MACROS_VERSION, [$1]), -1,
     [m4_fatal([pkg.m4 version $1 or higher is required but ]PKG_MACROS_VERSION[ found])])
 ])dnl PKG_PREREQ
 
-dnl PKG_PROG_PKG_CONFIG([MIN-VERSION])
-dnl ----------------------------------
+dnl PKG_PROG_PKG_CONFIG([MIN-VERSION], [ACTION-IF-NOT-FOUND])
+dnl ---------------------------------------------------------
 dnl Since: 0.16
 dnl
 dnl Search for the pkg-config tool and set the PKG_CONFIG variable to
@@ -55,6 +55,12 @@ dnl first found in the path. Checks that the version of pkg-config found
 dnl is at least MIN-VERSION. If MIN-VERSION is not specified, 0.9.0 is
 dnl used since that's the first version where most current features of
 dnl pkg-config existed.
+dnl
+dnl If pkg-config is not found or older than specified, it will result
+dnl in an empty PKG_CONFIG variable. To avoid widespread issues with
+dnl scripts not checking it, ACTION-IF-NOT-FOUND defaults to aborting.
+dnl You can specify [PKG_CONFIG=false] as an action instead, which would
+dnl result in pkg-config tests failing, but no bogus error messages.
 AC_DEFUN([PKG_PROG_PKG_CONFIG],
 [m4_pattern_forbid([^_?PKG_[A-Z_]+$])
 m4_pattern_allow([^PKG_CONFIG(_(PATH|LIBDIR|SYSROOT_DIR|ALLOW_SYSTEM_(CFLAGS|LIBS)))?$])
@@ -75,6 +81,9 @@ if test -n "$PKG_CONFIG"; then
 		AC_MSG_RESULT([no])
 		PKG_CONFIG=""
 	fi
+fi
+if test -z "$PKG_CONFIG"; then
+	m4_default([$2], [AC_MSG_ERROR([pkg-config not found])])
 fi[]dnl
 ])dnl PKG_PROG_PKG_CONFIG
 


kaniini (approved):

kaniini:
Merged, thanks!

eli-schwartz:
I think that this is a terrible proposal and the PR should be reverted and re-released.

It causes an immediate regression for the *common* use case of `pkg.m4`, which is to not use PKG_PROG_PKG_CONFIG at all, but use:

```m4
PKG_CHECK_MODULES([FOO], [foo], , [
    dnl fall back to probing the library itself
    AC_CHECK_HEADER([foo.h])
    AC_CHECK_LIB([foo]) 
])
```

All these existing, ***correctly written*** autoconf scripts, now become *invalid* as a result of this change. Making existing correctly-written projects break, is much much worse than permitting broken projects to remain broken.

Broken projects should simply fix their brokenness. In fact they aren't even "broken" at all, since "weird error messages through the run, which often confuse users" is still an error, and users are empowered to report a bug to the project in question.

kaniini:
Do you have an example of a package that has now been broken by this change?  It seemed reasonable to me, which is why I merged it.  We also haven’t seen any issue in Alpine, and 3.20 rebuild completed successfully with this patch installed.

eli-schwartz:
Well, like I said the entire point of the issue is that this breaks projects which support non-linux systems where pkg-config is not installed by default by abuild, or perhaps no one bothered to package it at all. Android sdk, commercial unix, you name it.

They are going to produce new distfiles on an updated linux development machine, it is going to "work", and then those projects are going to get angry bug reports from their users that the project no longer builds despite the promise that pkg-config is optional.

As an example:

I have recently been working on a MR to libtiff, a project that has optional components which are detected via autoconf-archive's AX_CHECK_GL macro. This is intended to work on systems where pkg-config isn't installed, but -lGL exists.

New versions of autoconf-archive (by which I mean improvements committed in 2013) support probing via pkg-config, and falling back to -lGL. But libtiff had checked their copy of a bunch of macros into git as m4/acinclude.m4. I proposed an MR to start depending on autoconf-archive (and, transitively, pkg-config!) installed when creating the dist tarball. With pkgconf 2.2.0, then simulating the *intended userbase* by `mv /usr/bin/pkg-config{,.bak}`:

```
checking for pkg-config... no
checking for gl... no
checking for library containing glBegin... -lGL
checking for GL/gl.h... yes
checking for compiling a minimal OpenGL program... yes
checking for linking a minimal OpenGL program... yes
checking for pkg-config... no
checking for glu... no
checking for library containing gluBeginCurve... -lGLU
checking for GL/glu.h... yes
checking for compiling a minimal OpenGL Utility (GLU) program... yes
checking for linking a minimal GLU program... yes
checking if GLU varargs tesselator is using non-standard form... no
checking for library containing glutMainLoop... -lglut
checking for GL/glut.h... yes
checking for compiling a minimal GLUT program... yes
checking for linking a minimal GLUT program... yes
```

With pkgconf 2.3.0:

```
checking for pkg-config... no
configure: error: pkg-config not found
```


The PR proposes that libtiff is "wrong", even though the only macro it uses is `AX_CHECK_GL`.

eli-schwartz:
It also means that pkg.m4 is now a big fat liar.

```
$ ./configure --help

[...]

  GL_CFLAGS   C compiler flags for GL, overriding configure script defaults
  GL_LIBS     Linker flags for GL, overriding configure script defaults

[...]
```

```
$ grep FAKE configure.ac
PKG_CHECK_MODULES([FAKE], [fake])

$ ./configure
[...]

configure: error: The pkg-config script could not be found or is too old.  Make sure it
is in your PATH or set the PKG_CONFIG environment variable to the full
path to pkg-config.

Alternatively, you may set the environment variables FAKE_CFLAGS
and FAKE_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

To get pkg-config, see <http://pkg-config.freedesktop.org/>.
```

The PKG_CHECK_MODULES macro has always been designed as something you can *brag about* to other projects: using it is idiomatic and provides extensive flexibility to your project, therefore people who are conservative about their build scripts and use AC_CHECK_LIBS or hand-rolled `if [ -n $PKG_CONFIG ]; then` for portability *should* be encouraged to use PKG_CHECK_MODULES because "Don't worry! pkg-config is optional, we are not shoving it down your throat. You can always specify the exact compile/link flags you need as environment variables, and PKG_CHECK_MODULES won't even check whether pkg-config is installed."

But you cannot use those settings because "configure: error: pkg-config not found". Unlike freedesktop's well designed error messages, this one doesn't even tell you what you should do about it.

eli-schwartz:
> Authors wishing to fallback to other methods can either return to the old behavior by specifying [:] as an action, or use [PKG_CONFIG=false] and call pkg-config macros normally, handling the fallback as any other failure.

You did not even bother to test this; it does not work.

```
$ PKG_CONFIG=false ./configure
checking pkg-config is at least version 0.9.0... no
configure: error: pkg-config not found
```

eli-schwartz:
@kaniini

> Do you have an example of a package that has now been broken by this change?

Can I assume providing an example wasn't helpful and this is being ignored instead?

Issue #352, opened 2024-04-05
Non-library dependency


ppisar:
I have a case where a library's public header includes a header from a different project so that application including the public header transitively includes the different project's header. But there is no need for the application to link to the different project's library.

My problem is that adding the other project to Requires of the library starts linking the application against the different project.

Concrete example is <https://github.com/rpm-software-management/librepo/issues/305>:

zck.pc contains "Libs: -lzck"
librepo.pc contains "Libs: glib-2.0"
librepo's /usr/include/librepo/downloadtarget.h includes "<zck.h>".

A librepo's user rightfully claims the compiling his application fails because including <librepo/downloadtarget.h> fails on missing zck.h.

If I change librepo.pc to "Libs: glib-2.0 zck", it will fix the user's problem. But "pkg-config --libs librepo" will produce "-lzck" (in addition to glib-2.0 link flags) because pkg-config thinks that every dependency is a library dependency. In this case it's not true, the application does not need linking to zck. The new librepo.pc would cause overlinking.

I can image these solutions:

1. zchunk project to deliver zck-headers.pc without "Libs:" in addition to zck.pc and librepo.pc to "Requires: zck-headers". This does not need changes in pkgconf, but it's not a generic solution. Similar problem can have any other project. I don't think projects should start deliver multiple pkg-config files just because somebody mights have this use case.
2. pkgconf to introduce a new non-library dependency which would suppress extending "pkg-config --libs". E.g. "Requires.headers: zck" in librepo.pc would add zck as a dependency of librepo and addedd zck's include path to "pkg-config --cflags librepo" output, but it wouldn't add "-lzck" to "pkg-config --libs librepo".
3. Do not change pkgconfig, neither zchunk. Add zck to librepo.pc Requires and claim that overlinking is a lesser problem then failing compilation.
4. Do not change anything and reject the user's request with a claim that pkg-config is about for tracking libraries, not headers and user needs the declare the dependency in other means than with pkg-config file.

What's your opinion?

dg0yt:
You should simply add the dependency to `Requires.private`. Other than one would expect, both `pkg-config` and `pkgconf` pull `cflags` from `Requires.private` even when not asking for `--static`. It has a test here:
https://github.com/pkgconf/pkgconf/blob/06120a8769aed87d50e914f87a6f9f67110cf16e/tests/requires.sh#L32-L38

[...]

Issue #300, opened 2023-06-20
Question: correct way to use Requires.private


freeekanayaka:
Hi,

I'm trying to understand how to correctly use the `Requires.private` property in a `.pc` file.

Let's say that `libA` internally depends on `libB` and that `libA.so` is dynamically linked to `libB.so` when `libA` is built as shared library.

My understanding is that `libA` should include `libB` in its `Requires.private` `.pc` file, is that right?

That would be mainly in order to support static builds.

Now, let's say that `my-app` is using `libA` and uses `PKG_CHECK_MODULES` in its `configure.ac` to detect the presence of `libA`. As far as I can see if `Requires.private` is used, the `configure` script of `my-app` will fail with something:

```
configure: error: Package requirements (libA >= 0.0.1) were not met:

Package 'libB', required by 'libA', not found
```

if the `.pc` file of `libB` is not installed on the system. This happens because the `pkg.m4` macros generates a command like:

```
/usr/bin/pkg-config --cflags "libA >= 0.0.1"
```

which fails with:

```
Package libB was not found in the pkg-config search path.
Perhaps you should add the directory containing `libB.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libB', required by 'libA', not found
```

Distributions like Debian ship `.pc` files in a separate `-dev` package, so the scenario in which the `.pc` file of `libA` is present, but the one of `libB` is not is actually the normal one, because a user building `my-app` should only need the the `libA-dev` package.

Should perhaps `Libs.private` be used instead? I'm confused about the proper way to do this.

Thanks

kaniini:
Requires.private is always considered for CFLAGS, because the module's headers may be referenced by the dependent (in fact, this would make a lot of sense, consider a module which builds on top of GTK, for example, it would need to reference the GTK headers).

Requires.private is only considered for LIBS if `--static` is also requested.

freeekanayaka:
> Requires.private is always considered for CFLAGS, because the module's headers may be referenced by the dependent (in fact, this would make a lot of sense, consider a module which builds on top of GTK, for example, it would need to reference the GTK headers).
> 
> Requires.private is only considered for LIBS if `--static` is also requested.

Thanks for your reply, I had missed it and I'm only reading it now. What if the modules headers are **not** referenced by the dependent? Is there a way to have CFLAGS behave in the same way as LIBS? I.e. make Requires.private only considered if --static is requested, **both** for LIBS **and** CFLAGS.

[...]

PR #353, opened 2024-04-08
doc: Requires.private is used for header inclusion

ppisar:
Document that "pkgconf --cflags" inherits paths for including headers from dependencies listed in Requires.private.

https://github.com/pkgconf/pkgconf/issues/300
https://github.com/pkgconf/pkgconf/issues/352


kaniini:
Merged, thanks!

PR #354, opened 2024-04-11
Fix typo in README.md

mokou:


dg0yt:
TBH the whole paragraph should be revisited. IMO there is no strong reason to generally discourage building with meson on non-Windows platforms. The meson build should be fully platform agnostic. I do understand that there may be bootstrapping issues with regard to meson. But that is a separate topic.

kaniini:
Merged, thanks!

PR #369, opened 2024-11-19
pkg.m4: Bump serial.  Use URL instead of obsolete FSF postal address.

jas4711:
Hi!

 I noticed you copied `pkg.m4` from pkg-config.  The file uses the old FSF postal address.

This pull request update it to use the latest GPLv2 recommended style with URL instead, the FSF has made this change in the GPLv2 license too: https://www.gnu.org/licenses/old-licenses/gpl-2.0.html

Also this bumps the serial which is necessary for all changes (should have been done with your earlier changes too).

What do you think?

Thanks,
/Simon

### DIFF
--- a/pkg.m4
+++ b/pkg.m4
@@ -1,5 +1,5 @@
 # pkg.m4 - Macros to locate and use pkg-config.   -*- Autoconf -*-
-# serial 12 (pkg-config-0.29.2)
+# serial 13 (pkgconf)
 
 dnl Copyright © 2004 Scott James Remnant <scott@netsplit.com>.
 dnl Copyright © 2012-2015 Dan Nicholson <dbn.lists@gmail.com>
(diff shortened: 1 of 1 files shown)


kaniini:
Sure, I will pull this in.

Longer term, I would prefer if `pkg.m4` is maintained by the autoconf maintainers (or that "autoconf archive" project).  If anyone knows how to facilitate that...

jas4711:
Thank you!

I was reading the contribution policy for autoconf-archives, and I don't think pkg.m4 belongs there.  The autoconf-archive is intended for custom macros with the AX_* namespace, written for the autoconf-archives specifically.  It is not intended as a repository for external autoconf M4 macros

Why don't you want to maintain it in pkgconf?  It is widely used.

kaniini:
The autoconf macros are specific to autoconf, and I have no interest in defining policy for how the macros are consumed.  This is why I would prefer that the macro library move to something more under the auspices of the autoconf project.

PR #386, opened 2025-03-20
 meson: add MSI build target 

elmarco:
It would be nice to provide a release MSI for Windows users.

Requires wixl & unoconv.

Tested on Fedora, with mingw64 cross-build.

### DIFF
--- a/libpkgconf/stdinc.h
+++ b/libpkgconf/stdinc.h
@@ -32,7 +32,7 @@
 # include <malloc.h>
 # define PATH_DEV_NULL	"nul"
 # ifdef _WIN64
-#  define SIZE_FMT_SPECIFIER	"%I64u"
+#  define SIZE_FMT_SPECIFIER	"%llu"
 # else
 #  define SIZE_FMT_SPECIFIER	"%u"
 # endif
--- a/meson.build
+++ b/meson.build
@@ -1,8 +1,8 @@
 project('pkgconf', 'c',
   version : '2.4.3',
   license : 'ISC',
-  meson_version : '>=0.49',
   default_options : ['c_std=c99'],
+  meson_version: '>=0.52',
 )
 
 cc = meson.get_compiler('c')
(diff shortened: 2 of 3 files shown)


kaniini:
I'll accept the patches, but as with everything else Windows-related, I don't have the infrastructure or time to support it officially.

libpkgconf/stdinc.h
```diff
@@ -32,7 +32,7 @@
 # include <malloc.h>
 # define PATH_DEV_NULL	"nul"
 # ifdef _WIN64
-#  define SIZE_FMT_SPECIFIER	"%I64u"
+#  define SIZE_FMT_SPECIFIER	"%llu"
```
kaniini, line 35: This fix is wrong.  We should use `%zu` on GNU C, but `%I64u` is the correct specifier for Microsoft C.
```suggestion
#  ifdef __GNUC__
#   define SIZE_FMT_SPECIFIER	"%zu"
#  else
#   define SIZE_FMT_SPECIFIER	"%I64u"
#  endif
```

kaniini:
WDYT @dg0yt?

dg0yt:
The PR delivers building an MSI for Windows on Linux. 
- It seems to be optional and to not interfere with MSVC. This is good.
- It would be nice if the MSI would also be created with MSVC.
- Does the MSI also bundle the runtime libs which are not from pkgconf (i.e. gcc, mingw)?

Disclaimer: I don't use MSVC locally...

Sidenote:
For ad-hoc usage including CI, it would be convenient to have a simple ZIP of the binaries.

elmarco:
>     * It would be nice if the MSI would also be created with MSVC.

How MSVC creates MSI?  How to even build pkgconf with MSVC?

> 
>     * Does the MSI also bundle the runtime libs which are not from pkgconf (i.e. gcc, mingw)?

The mingw build doesn't require extra dependencies. Thus the installed files (exe and dll) are enough.

> 
> 
> Disclaimer: I don't use MSVC locally...
> 
> Sidenote: For ad-hoc usage including CI, it would be convenient to have a simple ZIP of the binaries.

I tried to add "Upload Artifact" to the github workflow, it needs to be tested.

dg0yt:
> > ```
> > * It would be nice if the MSI would also be created with MSVC.
> > ```
> 
> How MSVC creates MSI? How to even build pkgconf with MSVC?

Sorry that my sentence was confusing. I meant building pkgconf with MSVC, not the MSI.

FTR meson supports MSVC, at least with the machine files generated in vcpkg.

> > ```
> > * Does the MSI also bundle the runtime libs which are not from pkgconf (i.e. gcc, mingw)?
> > ```
> 
> The mingw build doesn't require extra dependencies. Thus the installed files (exe and dll) are enough.

I would like to verify that. Checking a local mingw crossbuild with vcpkg on linux:
~~~
$ x86_64-w64-mingw32-objdump installed/x64-mingw-dynamic/tools/pkgconf/pkgconf.exe -x | grep DLL
 vma:            Hint    Time      Forward  DLL       First
        DLL Name: KERNEL32.dll
        DLL Name: msvcrt.dll
        DLL Name: libpkgconf-6.dll
~~~
Okay, this would be fine. (Probably because it is just C.)

> > Disclaimer: I don't use MSVC locally...
> > Sidenote: For ad-hoc usage including CI, it would be convenient to have a simple ZIP of the binaries.
> 
> I tried to add "Upload Artifact" to the github workflow, it needs to be tested.

I saw the GHA scripts, and I checked your fork before posting. But I didn't see actions enabled.

kaniini:
I merged this, but it seems to break CI, because `unoconv` is missing, and the 32-bit build package name is wrong.  Going to look into fixing both of these...

PR #392, opened 2025-04-03
Allow individual files to be a path to searched

### LINKED ISSUE #391: Add single .pc file to search space
For pkgconf-pypi we are setting PKG_CONFIG_PATH to allow pkgconf to be used to discover .pc files embedded in the python site-packages. We would like to add the python installation's .pc files to the list of discoverable packages, but sometimes those files are installed in a global location and we don't want to add those other global .pc files to the search path.

Is there a way to add single .pc files to the search path? If not, are you open to that change?

Reference: https://github.com/pypackaging-native/pkgconf-pypi/issues/69

virtuald:
Obviously needs tests and docs and whatever, but here's at least one implementation that fixes #391

### DIFF
--- a/libpkgconf/pkg.c
+++ b/libpkgconf/pkg.c
@@ -825,6 +825,48 @@ pkgconf_pkg_find_in_registry_key(pkgconf_client_t *client, HKEY hkey, const char
 }
 #endif
 
+static pkgconf_pkg_t *
+pkgconf_pkg_find_in_specific_envvar(pkgconf_client_t *client, const char *name)
+{
+	pkgconf_pkg_t *pkg = NULL;
+	char *envpath;
+
+	if ((envpath = getenv("PKG_CONFIG_EXTERNAL_FILES")) != NULL)
+	{
+		char search_suffix[PKGCONF_ITEM_SIZE] = {PKG_DIR_SEP_S, 0};
+		pkgconf_list_t plist = PKGCONF_LIST_INITIALIZER;
+		pkgconf_node_t *n;
+		FILE *f;
+
+		pkgconf_strlcat(search_suffix, name, sizeof search_suffix);
+		pkgconf_strlcat(search_suffix, PKG_CONFIG_EXT, sizeof search_suffix);
+
+		pkgconf_path_split(envpath, &plist, true);
+
+		PKGCONF_FOREACH_LIST_ENTRY(plist.head, n)
+		{
+			pkgconf_path_t *pnode = n->data;
+
+			PKGCONF_TRACE(client, "trying direct path: %s for %s", pnode->path, name);
+
+			if (!str_has_suffix(pnode->path, search_suffix))
+				continue;
+
+			if ((f = fopen(pnode->path, "r")) != NULL)
+			{
+				PKGCONF_TRACE(client, "found: %s", pnode->path);
+				pkg = pkgconf_pkg_new_from_file(client, pnode->path, f, 0);
+				if (pkg != NULL)
+					break;
+			}
+		}
+
+		pkgconf_path_free(&plist);
+	}
+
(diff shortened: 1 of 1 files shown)


libpkgconf/pkg.c
```diff
@@ -711,8 +711,29 @@ pkgconf_pkg_scan_dir(pkgconf_client_t *client, const char *path, void *data, pkg
 	pkgconf_pkg_t *outpkg = NULL;
 
 	dir = opendir(path);
-	if (dir == NULL)
+	if (dir == NULL) {
```
FFY00, line 714: Perhaps we want to check for `PKG_CONFIG_EXT` (`.pc`) here before trying to parse the file?

virtuald:
> Write tests first?

I'm not going to spend time to write tests if the idea isn't acceptable -- and that's why it's currently a draft PR. I did test it locally however.

dg0yt:
https://github.com/pkgconf/pkgconf/blob/1d37e711cefe83e6b5393b424af4c94da4e7e9d3/man/pkgconf.1#L179-L186

A "list of directories" is the interpretation shared with other implementations. IMO changing the (default) behavior only in `pkgconf` might be controversial when the new pattern is adopted outside the `pkgconf-pypi` universe.

In addition to the environment variables, there is also a `--path` command line option which takes a directory now. I assume the PR will change the behavior of what is now
~~~
$ ./installed/x64-linux/tools/pkgconf/pkgconf --with-path=installed/x64-linux/lib/pkgconfig/Imath.pc Imath --libs
Package Imath was not found in the pkg-config search path.
Perhaps you should add the directory containing `Imath.pc'
to the PKG_CONFIG_PATH environment variable
Package 'Imath' not found
~~~
into
~~~
$ ./installed/x64-linux/tools/pkgconf/pkgconf --with-path=installed/x64-linux/lib/pkgconfig/Imath.pc Imath --libs
-Linstalled/x64-linux/lib/pkgconfig/../../lib -lImath-3_1
~~~
which may look beneficial for this simple dependency but actually needs the *directory* for cases of transitive dependencies.

virtuald:
Good points. 

It seems like the only way to configure pkgconf is either the command line or environment variables. Command line would probably be annoying. What about a different environment variable then? `PKG_CONFIG_FILES`

kaniini:
`PKG_CONFIG_FILES` seems fine to me, but I would rather name the environment variable something like `PKG_CONFIG_EXTERNAL_FILES` so it makes it more clear what this proposed functionality is for.

I still am somewhat concerned about this feature, as it essentially works by preloading the package data into the pkgconf cache.  But we may choose to change the behavior of the cache in the future, for example to have it only cache the latest X items it has considered.  I think preloaded files should probably not be tied directly to our current caching behavior?

virtuald:
I've pushed a different implementation that adds support for `PKG_CONFIG_EXTERNAL_FILES` which modifies `pkgconf_pkg_find` instead.

One flaw with this implementation is that the packages are not shown in `--list-all`, but I note that loading from the Windows Registry doesn't seem to show them either (I haven't tried it, not on Windows) so that's probably not a big deal?

kaniini:
It's closer, but I think we want to explicitly track packages that we have preloaded.  This would allow us to fix `--list-all` for preloaded packages, just by fixing `pkgconf_scan_all()` I think.

kaniini:
In addition, right now I believe preloaded packages will leak.  If we track them, e.g. by adding a second linked list node to pkgconf_pkg_t, then we can clean them up when we deinit the pkgconf_client_t.

kaniini:
I implemented leak-free preloaded packages in latest git.

virtuald:
Thank you so much! Sorry I wasn't able to get this over the finish line.

PR #393, opened 2025-04-07
Fix installation instructions

adamency:
Add instruction about running the script `./autogen.sh` before running `./configure` as the former generates the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ which is not covered, please get in touch.
 pkgconf is basically compiled the same way any other autotools-based project is
 compiled:
 
+    $ ./autogen.sh
     $ ./configure
     $ make
     $ sudo make install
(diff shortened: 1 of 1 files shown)


adamency:
cc @kaniini

dg0yt:
Traditionally, running `autogen.sh` (or `autoreconf`) is a maintainer task for releases, and "dist" tarballs come with a pregenerated `configure`.
Here on GH, there are no such dist tarballs for releases, only on-the-fly tarballs for git tags.  
The README has a link to https://distfiles.ariadne.space/pkgconf/ with release tarballs.  
Maybe this needs to be elaborated a little bit more.

kaniini:
I think we will decline for the reasons @dg0yt outlined.  We don't want people using the github tarballs.  Sadly, there is not a way to disable them, which is frustrating.

PR #395, opened 2025-05-03
man/pkgconf.1: use conventional syntax for the synopsis line

ischwarze:
Hello,

OpenBSD <www.openbsd.org> just imported pkgconf(1) into the OpenBSD base system, see
https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/pkgconf/
The main reason for switching from our old implementation of pkg-config(1) written in Perl
https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/pkg-config/
to the more widely used and more actively developed pkgconf(1) was that our old pkg-config(1) was unbearably slow when dealing with projects using large numbers of modules.

My job in this context is getting the documentation = manual pages up to speed.
I'm the maintainer of the mandoc(1) documentation toolkit <mandoc.bsd.lv>
that is used by default by OpenBSD, Alpine, Void and Chimera Linux, FreeBSD, NetBSD, illumos and macOS and also optionally available on Arch, SUSE, Fedora, Gentoo, and Debian.
I'm also a developer in the GNU roff project, <www.gnu.org/software/groff>.

Rather than improving the manual pages only in the OpenBSD tree, my plan is to work upstream with you instead, which will likely result in large numbers of pull requests because the pkgconf(1) documentation is currently significantly below OpenBSD quality standards.  This is not intended as an insult, i'm not saying your documentation is bad, but merely as a statement of fact.  OpenBSD documentation quality standards are unusually high and unusually strict.

My plan is to start with very small patches, implementing one small improvement each, in the hope that they will be easy to verify and pull, or if need be, easy to reason about.  When we get to know each other better and i develop a feeling how you usually work, maybe patches can become slightly bigger over time.

Here is the first small patch:
 * In the synopsis line, avoid prose like "list of modules".  The conventional way to express a list of one or more arguments is to provide a placeholder for the first argument ("module"), then use three dots to indicate that more arguments of the same kind can optionally follow.  Here is a typical example: https://man.openbsd.org/file.1
 * Introduce the placeholder ".Ar module" such that it can later be used in the text, making the text more precise.  Such future usage is not yet included in this patch.
 * The "module" argument is *not* optional.  For normal use of the program, that argument is always required.  If you wonder about the very small number of options like --version and --help where the argument can be omitted, considering those in the synopsis would massively over-emphasize them.  If at all, the description of --version and similar options could later say that this option does not require an argument and ignores it if provided.  It is not usually possible to address every arcane special case in the synopsis, and trying to merely confuses users.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -39,7 +39,12 @@ This option is implied by many other options, but not by all.
 It can be overridden with
 .Fl -silence-errors .
 .It Fl -silence-errors
-Do not display any errors at all.
+Do not print any error, warning, or debugging messages at all.
+Overrides all of
+.Fl -debug ,
+.Fl -errors-to-stdout ,
+and
+.Fl -print-errors .
 .It Fl -debug
 Print some non-fatal warning messages to standard error output
 that would otherwise silently be ignored.
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -133,14 +133,14 @@ or only the include path
 flags, or only the compiler flags that are not include path flags,
 respectively.
 .It Fl -libs , Fl -libs-only-L , Fl -libs-only-l , Fl -libs-only-other
-Display either all linker flags, only
-.Fl L
-linker flags, only
-.Fl l
-linker flags or only linker flags that are not
-.Fl L
-or
-.Fl l .
+Print all linker flags required to link against the
+.Cm module ,
+or only the library path
+.Pq Fl L
+flags, or only the library
+.Pq Fl l
+flags, or only the linker flags that are neither library path
+nor library flags, respectively.
 .It Fl -keep-system-cflags , Fl -keep-system-libs
 Keep CFLAGS or linker flag fragments that would be filtered due to being
 included by default in the compiler.
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -118,10 +118,12 @@ of a solution.
 .Sh QUERY-SPECIFIC OPTIONS
 .Bl -tag -width indent
 .It Fl -cflags , Fl -cflags-only-I , Fl -cflags-only-other
-Display either all CFLAGS, only
-.Fl I
-CFLAGS or only CFLAGS that are not
-.Fl I .
+Print all compiler flags required to compile against the
+.Cm module ,
+or only the include path
+.Pq Fl I
+flags, or only the compiler flags that are not include path flags,
+respectively.
 .It Fl -libs , Fl -libs-only-L , Fl -libs-only-l , Fl -libs-only-other
 Display either all linker flags, only
 .Fl L
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -16,7 +16,7 @@
 .Sh SYNOPSIS
 .Nm
 .Op Ar options
-.Op Ar list of modules
+.Ar module ...
 .Sh DESCRIPTION
 .Nm
 is a program which helps to configure compiler and linker flags for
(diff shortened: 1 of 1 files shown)


ischwarze:
Here is the rationale for the second commit, [c209b0d](https://github.com/pkgconf/pkgconf/pull/395/commits/c209b0d73d0ee1db327a937cf64f4058c3f43cb6).

 * Minor unification of terminology: sometimes saying "display" and sometimes saying "print" for the same meaning is confusing.  The most precise term would be "print to standard output", but repeating "to standard output" over and over again would probably be excessive.
 * Avoid jargon like "CFLAGS", use the plain English term "compiler flags" instead.  Besides, mentioning the string "CFLAGS" isn't needed as it already appears in the option name.
 * Add missing information - which compiler flags?  For maximum precision, reference the ".Ar module" argument as explained for the first patch.
 * Use a proper English term for the -I flags.
 * Use the Oxford comma in the enumeration with "or"; it is popular in formal technical writing.
 * Add the word "respectively" at the end ro make it more explicit what the "or"s mean.

In OpenBSD, we usually avoid disussing multiple command line options in the same list entry, in particular when they differ in purpose or semantics.  So i'd (slightly) prefer splitting this into three separate list entries, but do not feel very strongly about it.  What do you think?

ischwarze:
Rationale for the third commit, [aa69ff8](https://github.com/pkgconf/pkgconf/pull/395/commits/aa69ff8b41ab0eb0c18efd214055516dfe55afc1):

 * The --debug option is compiled in by default and provides non-trivial functionality, so it needs to be documented.
 * The normal way to document command line options is in alphabetical order, to make it easier for users to find the option they are looking for. Here, the order of the options in the manual page looks completely random.  That should almost certainly be fixed, but i didn't start the sorting effort yet because having your feedback on the general idea first would make sense.
  * Consequently, the place where to put --debug is quite arbitrary for now.
  * The fact that --debug prints warning messages to standard output is almost certainly a bug.  In Unix-like operating systems, such messages have to be printed to standard error output.  However, right now, i'm only trying to fix the documentation, not the code, so for now, let's document the current state of affairs.
  * The user interface design also looks dubious in so far as the --debug option not only prints debugging messages, but is also required to get various warning messages.  Again, for now, just document the current behaviour, however dubious it may seem.

ischwarze:
Here is the rationale for the sixth commit, [94beccc](https://github.com/pkgconf/pkgconf/pull/395/commits/94becccb744330f3f622b060e8d46be2d0e37280):

* The option --errors-to-stdout does not only affect errors, but also warnings and debugging.  Add the missing information.
* The "stdout" stream is called "standard output" in almost all manual pages, so use the normal wording.  Using the word "main" would be highly unusual and could easily mislead users into thinking that pkconf(1) sets up another, additional, auxiliary output stream.  While the word "stream" is unlikely to cause harm, it is so unusual that it seems better to leave it out.
* The "stderr" stream is called "standard error output" in the majority of manual pages.  Some manual pages say just "standard error" and leave out the word "output", but that feels slightly sloppy.  Again, no need for the word "stream".

Sorry for misreading the code earlier.  Please disregard the spurious complaint above, messages do go to stderr as they should.

ischwarze:
Regarding the seventh commit, [b44bce1](https://github.com/pkgconf/pkgconf/pull/395/commits/b44bce162f7bb0772a5bad69ca41040cf73484e9):

A more precise description would be desirable.  Currently, the description of the error handling is not particularly helpful from the user persepective.  On the other hand, adding "implies --print-errors" to the description of more than a dozen other options does not feel like a sustainable direction either.

Ideally, the manual page ought to exhaustively describe which messages are printed under which conditions, and the rules ought to be *simple* - the fewer options, the better.

kaniini:
Patches merged!

PR #397, opened 2025-05-04
man/pkgconf.1: sort options list; no text or markup change

ischwarze:
Rationale:
 * In practically every section 1 manual page, the list of command line options is among the most prominent elements of the DESCRIPTION, and in most cases, *the* main component of the DESCRIPTION.  The options list should never be moved out of the DESCRIPTION into some other section(s) because that violates the expectations of experienced users.
 * For introducing the options list, use the conventional sentence "The options are as follows:".
 * Using custom section names (custom .Sh arguments) in manual pages is generally discouraged, because uniform naming of sections makes it easier for users to find their way around and get used to how manual pages are laid out.  Exceptions are sometimes justified when the running text of the DESCRIPTION becomes extremely long and complicated (like it might, for example, in a sh(1) manual page), and in that case, splitting custom sections out of the DESCRIPTION sometimes makes sense.  However, so far, the DESCRIPTION of pkgconf(1) - apart from the options list, which must NEVER be split out, not even in an extremely complex manual page like sh(1) - is just three lines long, so custom sections are not useful here.

The correctness of this patch can be veriefied as follows:

  sort pkgconf.1 > before.1
  patch < this.patch
  sort pkgconf.1 > after.1
  diff -u before.1 after.1

The earlier we do this, the better for developers who run "git annotate" down the line.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -23,28 +23,22 @@ is a program which helps to configure compiler and linker flags for
 development libraries.
 This allows build systems to detect other dependencies and use them with the
 system toolchain.
-.Sh GENERAL OPTIONS
+.Pp
+The
+.Ar options
+are as follows:
 .Bl -tag -width indent
-.It Fl -version
-Display the supported pkg-config version and exit.
 .It Fl -atleast-pkgconfig-version Ns = Ns Ar VERSION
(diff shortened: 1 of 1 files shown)


kaniini:
Merged, thanks!

PR #398, opened 2025-05-05
man/pkgconf.1: improve the description of --atleast-pkgconfig-version

ischwarze:
Rationale for the first patch:
 * Mark up VERSION with .Ar in the text body such that it becomes clear it refers to the option argument.
 * For precision and clarity, consistently talk about version numbers rather than merely "versions".
 * Precisely state what the check is (greater than) instead of using the vague wording "support".
 * Make it clearer that the version number in question is the version number of the program itself, by using the .Nm macro, by avoiding the confusing string "pkg-config" (that's not what the program in this package is called), and by adding the word "program".
 * Avoid the word "we"; in manual pages, using the first and second persons is generally frowned upon.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -281,6 +281,8 @@ If set, follow the sysroot prefixing rules that freedesktop.org pkg-config uses.
 .It Va DESTDIR
 If set to PKG_CONFIG_SYSROOT_DIR, assume that PKG_CONFIG_FDO_SYSROOT_RULES is set.
 .El
+.Sh EXIT STATUS
+.Ex -std
 .Sh EXAMPLES
 Displaying the CFLAGS of a package:
 .Dl $ pkgconf --cflags foo


ischwarze:
Rationale for the second patch, [8fc2107](https://github.com/pkgconf/pkgconf/pull/398/commits/8fc2107159421df43fec5492d411c8d4a3f3c01c):
 * s/result/exit status/ for clarity and to adhere to standard terminology.
 * s/was/is/ because past tense is rarely needed in manual pages.
 * Mark up "module" with .Ar to make it clear that the word refers to the command line arguments.
 * Document the somewhat unusual user interface design where this option is the default, cannot be disabled, and yet does not guarantee that all arguments are inspected, not even when --exists is specified explicitly.

ischwarze:
Rationale for the third patch, [244c4da](https://github.com/pkgconf/pkgconf/pull/398/commits/244c4da9442d95cc73af3a9e729a0dc7ad249da3):

 * Avoid the imprecise word "relevant", it only makes the user wonder what is relevant and what isn't.
 * Mark up keywords in configuration files (like Provides) with .Ic.
 * Describe precisely which modules are processed.
 * Explicitly describe the Name=Version output line.
 * Describe precisely how Provides are handled.
 * Mention options implied and overridden by this option.

ischwarze:
Rationale for commit [a1a1809](https://github.com/pkgconf/pkgconf/pull/398/commits/a1a180924e0ff83fb8188be3a84158e3d8f3040d):

* Use the word "respectively" to make it clearer how the two options relate to the two properties.
* Mention that the output follows dependency list format.
* Mention that printing goes to standard output.
* Mention what these options imply and override.

PR #399, opened 2025-05-06
man/pkgconf.1: do not use all caps for the arguments of .Ar macros

ischwarze:
Very little text change, only minimal wording and markup improvements in the immediate vicinity.

Rationale:
In a manual page, it would be highly unusual to use all caps for the placeholders for option arguments.  In general, all caps harms accessibility because screen readers often pronounce the letters of such words one-by-one rather than pronouncing the whole word.  Consequently, unnecessary all caps is better avoided.

Additionally, there is no need for such pedestrian markup.  The .Ar macro already results in italic font, or on terminals that don't provide an italic font, it results in underlining.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -123,7 +123,7 @@ flags, or only the linker flags that are neither library path
 nor library flags, respectively.
 .It Fl -list-all
 Walk all directories listed in the
-.Va PKG_CONFIG_PATH
+.Ev PKG_CONFIG_PATH
 environmental variable and display information on packages which have registered
 information there.
 .It Fl -max-version Ns = Ns Ar version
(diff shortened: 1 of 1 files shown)


ischwarze:
Remark regarding commit [3ebcf00](https://github.com/pkgconf/pkgconf/pull/399/commits/3ebcf001f1dfff167c279e3d75ca38c80d402810):

The old text gave the misleading impression that an error would be raised if *any* module violated the respective condition.  In fact, *all* modules have to violate the condition for an error to be raised.  I verified that both by testing and by code inspection.

ischwarze:
Regarding commit [df34b89](https://github.com/pkgconf/pkgconf/pull/399/commits/df34b89fec91bef467ee41965fddd67630c648e6):

The following information was missing:

 * What happens with multiple "module" arguments.
 * The file descriptor the version numbers are printed to.
 * Implied and overridden options.
 * The --verbose option was not mentioned at all.

ischwarze:
Regarding commit [df5ccd1](https://github.com/pkgconf/pkgconf/pull/399/commits/df5ccd17b90d2385112c393ba78893a4ce2898e4):

 * I cannot see any evidence in the code how this option could possibly return more than one filename.  The code clearly only processes the first module, and then only one filename is extracted from that module.  So i believe that the current use of the plural "filenames" is a documentation bug.
 * For "pc", use the .Xr rather than the .Sq macro.
 * Avoid the confusing term "dependency set".  The noun "set" appears nowhere else in the pkgconf(1) or pc(5) manual pages, and it is unclear what it is supposed to mean.
 * Qualify the pathname that is printed as "absolute".
 * Say "print to standard output", not "display".
 * Mention the early exit behaviour of this option.

ischwarze:
Regarding commit [1d479be](https://github.com/pkgconf/pkgconf/pull/399/commits/1d479be94e9a0d08b6f53e1b3ca192f8d8731a81):

In manual pages, it is conventional to sort ENVIRONMENT sections alphabetically.

To mark up the names of environment variables, .Ev is used rather than .Va; the latter is better suited to variable names in compiled languages.

No text change - this patch can be verified as follows:

  $ sort man/pkgconf.1 > before.1
  $ patch < this_patch
  $ sort man/pkgconf.1 > after.1
  $ diff -u before.1 after.1

kaniini:
Merged, thanks!

PR #400, opened 2025-05-08
man/pkgconf.1: properly document the module search path

ischwarze:
Specifically, document --env-only, --list-all, --with-path, PKG_CONFIG_LIBDIR, and PKG_CONFIG_PATH, all of which were already mentioned, but very little was said about them, and most of what *was* said amounted to lies.

Unfortunately, the description ends up being rather lengthy and likely somewhat confusing for most readers, simply because the user interface design is a horrific mess and extremely counter-intuitive in several respects.  In particular, it is riddled with the following, highly surprising quirks:

* There are two environment variables serving a very similar purpose, but exhibiting several subtle semantic differences.
* One of the variables gets prepended to the system defaults, while the other overrides the system defaults.
* One of the variables can be supplemented with a command line option, while the other variable *overrides* the same command line option - elsewhere, environment variables overriding command line options are almost unheard of.
* The overriding variable has *lower* priority than the variable that can be combined with the system defaults.
* Both variables accept colon-separated lists of directories, but the command line option only accepts a single directory.

With such a quirky interface, writing documentation that is simple and concise is unfortunately not feasible.  The best one can do is optimize the description for being precise, detailed, and explicit, at the expense of becoming somewhat lengthy.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -84,7 +84,13 @@ Print the requested values as variable declarations in a similar format as the
 .Xr env 1
 command.
 .It Fl -env-only
-Learn about pkgconf's configuration strictly from environmental variables.
+Initialize the module search path from
+.Fl -with-path
+and
+.Ev PKG_CONFIG_PATH
+only, ignoring
+.Ev PKG_CONFIG_LIBDIR
+and the compiled-in default directories.
 .It Fl -errors-to-stdout
 Print all error, warning, and debugging messages to standard output
 instead of to standard error output.
@@ -122,10 +128,17 @@ flags, or only the library
 flags, or only the linker flags that are neither library path
 nor library flags, respectively.
 .It Fl -list-all
-Walk all directories listed in the
-.Ev PKG_CONFIG_PATH
-environmental variable and display information on packages which have registered
-information there.
+Walk the module search path in the order of descending priority.
+For each
+.Xr pc 5
+file found, print one line to standard output,
+containing the basename of the file without the extension, the
+.Ic Name
+property, a dash
+.Pq Sq \- ,
+and the
+.Ic Description
+property.
 .It Fl -max-version Ns = Ns Ar version
 Exit with error if the version number of each
 .Ar module
@@ -276,12 +289,13 @@ Print the version number of the
 program to standard output and exit.
 Most other options and all command line arguments are ignored.
 .It Fl -with-path Ns = Ns Ar path
-Add a new module search
+Prepend the directory
 .Ar path
-to
-.Nm Ap s
-dependency resolver.
-Paths added in this way are given preference before other paths.
+to the module search path,
+giving it priority over all other directories including those from
+.Ev PKG_CONFIG_PATH
+and
+.Ev PKG_CONFIG_LIBDIR .
 .El
 .Sh ENVIRONMENT
 .Bl -tag -width indent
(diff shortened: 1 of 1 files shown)


ischwarze:
Here are some examples of the most blatant lies fixed by the commit:
The old text lied that
 * --env-only restricted all configuration to environment variables, while actually, it is only concerned with the module search path, does *not* disable many configuration options that can be given on the command line, does *not* disable some configuration stuff that is compiled in, but on the other hand even knocks out one environment variable
 * --list-all only walked PKG_CONFIG_PATH
 * --with-path enjoyed preference, while actually, it is used as the lowest priority fallback, if at all, and PKG_CONFIG_LIBDIR even completely disables it
 * PKG_CONFIG_LIBDIR were "primary" while it's actually lower priority than PKG_CONFIG_PATH
 * PKG_CONFIG_PATH were "secondary" while in truth, it's always the highest priority source

I must admit that i have rarely seen documentation that was so outragously wrong in so many respects.
I really wonder how something like this could happen in the first place...

Or did i totally misread the code, and has all my testing been invalid?
If that should be the case, i apologize for getting this wrong.  :-/

kaniini:
I think there's probably many areas where the documentation isn't fully aligned with the current state of affairs.

Much of the manual pages were written about 10 years ago and pkgconf was a much different tool back then, for example the behavior of the solver was entirely different (since we had to redesign it from scratch to scale to these crazy DAGs that Google loves to throw at it).  Because of that, we've revisited many design decisions in the original pkg-config and decided to change them to be more sane over the years.

Another example is `PKG_CONFIG_LIBDIR` vs `PKG_CONFIG_PATH`.  In reality, there is only a single search list, which is initialized with `PKG_CONFIG_PATH` first (or the compiled in default), and then `PKG_CONFIG_LIBDIR` is appended.  The behavior you describe with `--with-path` being clobbered by `PKG_CONFIG_LIBDIR` is a bug and we should fix that bug instead of documenting it IMO.

I think part of the issue is that over the years the documentation has somewhat bitrotted because writing mdoc is ... a chore.

kaniini (changes requested):
Other than the `--with-path` fix I just committed, the rest of this seems basically aligned with reality.  Lets align the `--with-path` discussion with the corrected behavior, and then we can merge this batch.

ischwarze:
Hello Ariadne, thanks for explaining the background, i think i now understand better what is going on.

I'm currently working on rebasing my work on top of your work, then modifying it to reflect the intended and now implemented reality (which will also make the text slightly simpler and shorter), then force-push such that you can afterwards pull.

Since you said "there is only a single search list", it might make sense to intoduce a term designating that list, then use that term consistently throughout.  That might also help making the text shorter and simpler.  I lean towards calling it "the module search path", but i'm of course open to picking a different name that you like better.

One tiny detail - you said:

  "initialized with PKG_CONFIG_PATH first (or the compiled in default), and then PKG_CONFIG_LIBDIR is appended"

That statement seems incorrect to me.  If i read the code correctly, this would be nearer the truth:

  initialized with PKG_CONFIG_PATH first, and then PKG_CONFIG_LIBDIR (or the compiled in default) is appended

Right?

kaniini:
Yeah, you're right.  The default search list is replaced by PKG_CONFIG_LIBDIR if set in the environment.

kaniini:
Merged!

Issue #364, opened 2024-09-10
Libs.private is ignored on Windows causing Rust static libs to fail linking


amyspark:
Hi,

Given the following pc module,

```
prefix=C:/Users/Amalia/.cargo
exec_prefix=${prefix}
libdir=${prefix}/lib
includedir=${prefix}/include

Name: example_project
Description: 
Version: 0.1.0
Libs: -L${libdir} -lexample-project
Cflags: -I${includedir}/example-project-0.1
Libs.private: -lkernel32 -ladvapi32 -lntdll -luserenv -lws2_32 
```

and pkgconf built with commit a88c0d962a987c62d98ede5a738e37ec71005cbd, I get the following output:

<details><summary> `pkg-config --libs --static example_project  `</summary>

```
❯ pkg-config --libs --static example_project 
../libpkgconf/client.c:528 [pkgconf_client_set_warn_handler]: installing default warn handler
../libpkgconf/client.c:257 [pkgconf_client_set_sysroot_dir]: set sysroot_dir to: <default>
../libpkgconf/tuple.c:251 [pkgconf_tuple_add]: adding tuple to @00007FF6CC06B618: pc_sysrootdir => / (parsed? 0)
../libpkgconf/client.c:301 [pkgconf_client_set_buildroot_dir]: set buildroot_dir to: <default>
../libpkgconf/tuple.c:251 [pkgconf_tuple_add]: adding tuple to @00007FF6CC06B618: pc_top_builddir => $(top_builddir) (parsed? 0)
../libpkgconf/client.c:489 [pkgconf_client_set_prefix_varname]: set prefix_varname to: prefix
../libpkgconf/client.c:142 [pkgconf_client_init]: initialized client @00007FF6CC06B5D0
../libpkgconf/client.c:39 [trace_path_list]: filtered library paths:
../libpkgconf/client.c:44 [trace_path_list]:   - 'C:/Users/Amalia/.cargo/lib'
../libpkgconf/client.c:39 [trace_path_list]: filtered include paths:
../libpkgconf/client.c:44 [trace_path_list]:   - 'C:/Users/Amalia/.cargo/include'
../libpkgconf/dependency.c:115 [add_or_replace_dependency_node]: added dependency [example_project] to list @000000F70AB5EE60; flags=0
../libpkgconf/dependency.c:232 [pkgconf_dependency_ref]: example_project refcount@000002A3ED58A7B0: 1
[...]

PR #410, opened 2025-06-12
personality.c: _WIN32 does not want_default_static

dg0yt:
Discussed in https://github.com/pkgconf/pkgconf/issues/364.
Applied in msys2's mingw packages, https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-pkgconf/0004-no-default-static.patch.
(Somewhat) tested in vcpkg.


dg0yt:
@kaniini Will this be in the coming release?

Issue #413, opened 2025-06-21
Weird output since 2.5.0 under cygwin


lazka:
2.4.3:

```
$ pkgconf libxml-2.0 --cflags
-I/usr/include/libxml2
```

2.5.0:

```
$ pkgconf libxml-2.0 --cflags
Package _tuple_add]: was not found in the pkg-config search path.
Perhaps you should add the directory containing `_tuple_add]:.pc'
to the PKG_CONFIG_PATH environment variable
Package '_tuple_add]:', required by 'libxml-2.0', not found
Package 'adding', required by 'libxml-2.0', not found
Package 'tuplc', required by 'libxml-2.0', not found
```

Bisected to 205bc7bf4266945f02d7313ccdd917074a91b09e and reverting that commit fixes things.

If there is anything I should test, let me know.

jonaski:
I'm getting strange issues on Windows with 2.5.0:

```
C:\Data\Projects\strawberry\strawberry-msvc>c:\strawberry_msvc_x86_64_debug\bin\pkgconf libpng --libs
Package ²²²²x was not found in the pkg-config search path.
Perhaps you should add the directory containing `²²²²x.pc'
to the PKG_CONFIG_PATH environment variable
Package '²²²²x', required by 'zlib', not found

```

jonaski:
Reverting https://github.com/pkgconf/pkgconf/commit/205bc7bf4266945f02d7313ccdd917074a91b09e does not help for me.

But the issue looks to be with zlib.pc, which is listed as ``Requires.private: zlib``,
if I remove the line ``Requires:`` line from zlib.pc, it works.

zlib.c looks like this:

```
prefix=C:/strawberry_msvc_x86_64_debug
exec_prefix=C:/strawberry_msvc_x86_64_debug
libdir=C:/strawberry_msvc_x86_64_debug/lib
sharedlibdir=C:/strawberry_msvc_x86_64_debug/lib
includedir=C:/strawberry_msvc_x86_64_debug/include

Name: zlib
Description: zlib compression library
Version: 1.3.1

Requires:
Libs: -L${libdir} -L${sharedlibdir} -lzlibd
Cflags: -I${includedir}

```

jonaski:
There are lot's of .pc files with empty ``Requires:`` lines:
```
$ grep '^Requires:\s*$' /c/strawberry_msvc_x86_64_debug/lib/pkgconfig/*.pc
[...]

Issue #406, opened 2025-05-30
Improve windows support


kaniini:
I would like to just drop all of the Windows-specific hacks and figure out what we actually need verses what we don't need.

As an example, pkgconf supports augmenting default PKG_CONFIG_PATH using the Windows registry, but there is no evidence anyone actually uses this feature, which probably means it's not used.  I implemented it because pkg-config originally did, but they wound up dropping the feature later on.

Another example is the `want_default_static` and `want_default_pure` flags being enabled.  Do we actually need these?

We should also try to make the tests work on Windows, if we can.  I don't know if `kyua` supports Windows yet though.

cc @dg0yt @elmarco and anyone else interested in pkgconf on Windows.

dg0yt:
I support removing Window-specific hacks.

What remains useful, and is used, is the `--msvc-syntax` command line option.

> We should also try to make the tests work on Windows, if we can. I don't know if `kyua` supports Windows yet though.

`kyua` is a barrier for testing *on* Windows. It isn't even packaged in MSYS2.

Anything which needs a MSYS enviroment (read: `fork` etc.) in Windows will be a slow.
IMO options are: moving tests into a compiled binary, or using python (because meson already requires python.)

dg0yt:
Wine might add polyfill which a native Windows environment lacks. (Filepaths?)
And if someone wants to analyse problems or make contributions in a native Windows environment, it would be useful to have the chance to run the test suite in the native environment. This also avoids a barrier for adding new tests.

kaniini:
New harness is now building and running on Windows.  Now time to start porting the tests...

vtorri:
> Maybe we can replace `kuya` for something more cross-platform? I only know `GTest` but any other cross-platform testing lib will work.

`lexbor` is a project which has implemented a small unit test framework in C99 : https://github.com/lexbor/lexbor/tree/master/test/unit

in case it helps

kaniini:
There is already a new test suite which does exactly what I want the test suite to do, I have been porting tests to it.

moi15moi:
Currently, the installer only works with x86 and x64 (https://github.com/pkgconf/pkgconf/blob/21841e3829cdd9b5bf33dfa700b4d2572757b176/meson.build#L213).
It would be nice to also support arm.

kaniini:
Tests are now building and passing on MSVC.  I also cleaned up the warnings and disabled the ones that don't make sense in a portable codebase.

moi15moi:
[...]

PR #414, opened 2025-06-23
dependency: Add check for zero string

jonaski:
Fixes strange pkgconf results on Windows when there are empty ``Requires:`` lines in the .pc files.

Possible fix for #413

### DIFF
--- a/libpkgconf/dependency.c
+++ b/libpkgconf/dependency.c
@@ -318,6 +318,9 @@ pkgconf_dependency_parse_str(pkgconf_client_t *client, pkgconf_list_t *deplist_h
 
 	memset(cmpname, '\0', sizeof cmpname);
 
+	if (strlen(depends) == 0)
+		return;
+
 	buf_sz = strlen(depends) * 2;
 	buf = calloc(1, buf_sz);
 	if (buf == NULL)


libpkgconf/dependency.c
```diff
@@ -318,6 +318,9 @@ pkgconf_dependency_parse_str(pkgconf_client_t *client, pkgconf_list_t *deplist_h
 
 	memset(cmpname, '\0', sizeof cmpname);
 
+	if (strlen(depends) == 0)
+		return;
+
 	buf_sz = strlen(depends) * 2;
 	buf = calloc(1, buf_sz);
```
dg0yt, line 325: So this is the original problem. When `buf_sz` is zero, "the behavior is implementation defined (null pointer may be returned, or some non-null pointer may be returned that may not be used to access storage)."
https://en.cppreference.com/w/c/memory/calloc

dg0yt: And now we are again at the point that `pkgconf` lacks running tests on Windows/MinGW.

dg0yt: FTR running tests on Windows is a subissue of #406.

PR #429, opened 2025-10-10
Add SPDX Lite profile 3.0 generator

illuusio:
Add SPDX Lite profile 3.0 generator named spdxtool to pkdconf. It outputs JSON-LD RDF tree and does only depend pkgconf and pkgconf .pc file to generate SBOM.

It superseeds bomtool which is SPDX 2.x generator name bomtool. But as SPDX version 2.x is still widely used new tool was created instead of upgrading old one.

Implementation does not rely out of tree JSON implementation but implements own very bare bones JSON-LD serializer.

It's still bit WIP but works as expected.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -125,6 +125,16 @@ EXTRA_DIST =	pkg.m4 \
 		tests/lib1/cflags-libs-private-c.pc \
 		tests/lib1/truncated.pc \
 		tests/lib1/c-comment.pc \
+		tests/lib-sbom/meta_package.pc \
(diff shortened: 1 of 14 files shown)


kaniini:
This is pretty impressive!  At the moment I am tied up with the Alpine 3.23 release, but I will try to get this reviewed ASAP.

illuusio:
Thank you for kind words. I'll update code to go thru tests and get `Source` working before that.

kaniini:
Let's go ahead and merge this, we can improve on it in future PRs

PR #438, opened 2026-01-13
Add buffer copy and contains byte

Elizafox:
This PR adds `pkgconf_buffer_copy` and `pkgconf_buffer_contains_byte`. They do as the functions describe: copy a buffer, or check if a buffer contains a byte (using `strchr`).

This will be used soon for porting the tuple code to use buffers.

### DIFF
--- a/libpkgconf/libpkgconf.h
+++ b/libpkgconf/libpkgconf.h
@@ -499,7 +499,7 @@ static inline char *pkgconf_buffer_freeze(pkgconf_buffer_t *buffer) {
 	return out;
 }
 
-static inline void pkgconf_buffer_copy(pkgconf_buffer_t* buffer, pkgconf_buffer_t* newptr)
+static inline void pkgconf_buffer_copy(pkgconf_buffer_t *buffer, pkgconf_buffer_t *newptr)
 {
     pkgconf_buffer_reset(newptr);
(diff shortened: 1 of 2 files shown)


libpkgconf/libpkgconf.h
```diff
@@ -460,6 +460,7 @@ PKGCONF_API void pkgconf_buffer_fputs(pkgconf_buffer_t *buffer, FILE *out);
 PKGCONF_API void pkgconf_buffer_vjoin(pkgconf_buffer_t *buffer, char delim, va_list va);
 PKGCONF_API void pkgconf_buffer_join(pkgconf_buffer_t *buffer, char delim, ...);
 PKGCONF_API bool pkgconf_buffer_contains(const pkgconf_buffer_t *haystack, const pkgconf_buffer_t *needle);
+PKGCONF_API bool pkgconf_buffer_contains_byte(const pkgconf_buffer_t* haystack, char needle);
```
kaniini:
```suggestion
PKGCONF_API bool pkgconf_buffer_contains_byte(const pkgconf_buffer_t *haystack, char needle);
```

kaniini (changes requested):
style + compile fixes...

kaniini (approved):

PR #444, opened 2026-01-21
spdxtool: fix the build on FreeBSD

khorben:
This avoids the warnings "no previous prototype for function" and "function declaration without a prototype is deprecated", which are promoted as errors by default when building pkgconf imported within FreeBSD's src repository.

Sponsored by:	The FreeBSD Foundation

### DIFF
--- a/cli/spdxtool/serialize.c
+++ b/cli/spdxtool/serialize.c
@@ -12,6 +12,7 @@
 #include <string.h>
 #include "util.h"
 #include "core.h"
+#include "serialize.h"
 
 static inline void
 serialize_add_indent(pkgconf_buffer_t *buffer, unsigned int level)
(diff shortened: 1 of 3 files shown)


khorben:
It's because we are not building the code in the same context.

Building software within FreeBSD's src repository defaults to:
```
-Wno-format-zero-length -fstack-protector-strong -Wsystem-headers -Werror -Wall -Wno-format-y2k -W -Wno-unused-parameter -Wstrict-prototypes -Wmissing-prototypes -Wpointer-arith -Wreturn-type -Wcast-qual -Wwrite-strings -Wswitch -Wshadow -Wunused-parameter -Wcast-align -Wchar-subscripts -Wnested-externs -Wold-style-definition -Wno-pointer-sign -Wdate-time -Wmissing-variable-declarations -Wthread-safety -Wno-empty-body -Wno-string-plus-int -Wno-unused-const-variable -Wno-error=unused-but-set-parameter -Wno-error=cast-function-type-mismatch -Qunused-arguments
```

The default compiler settings are not as strict, even when building packages from the ports.

kaniini (approved):

PR #456, opened 2026-02-27
spdxtool: fix iteration in spdxtool_core_spdx_document_free

Elizafox:
When freeing a list, it is best to use `PKGCONF_FOREACH_LIST_ENTRY_SAFE`.

In addition, most of these checks for NULL are unnecessary; `free(3)` already performs these checks.

This also fixes some fields not being properly set to `NULL`. Not super necessary but good hygiene.

No functional change intended by this pull request.

### DIFF
--- a/cli/spdxtool/core.c
+++ b/cli/spdxtool/core.c
@@ -295,92 +295,35 @@ spdxtool_core_spdx_document_new(pkgconf_client_t *client, char *spdx_id, char *c
 void
 spdxtool_core_spdx_document_free(spdxtool_core_spdx_document_t *spdx_struct)
 {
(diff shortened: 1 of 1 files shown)


cli/spdxtool/core.c
```diff
@@ -295,93 +295,56 @@ spdxtool_core_spdx_document_new(pkgconf_client_t *client, char *spdx_id, char *c
...
-		iter_last = iter;
+		free(iter);
 	}
 
-	if(iter_last)
-	{
-		free(iter_last);
-		iter_last = NULL;
-	}
+	spdx_struct->licenses.head = NULL;
```
kaniini: why null something that is being freed?

PR #475, opened 2026-03-25
spdxtool: Add homepage handling

illuusio:
Add URL-keyword handling to spdxtool. If URL is available add software_homePage with correct  information if not add empty value to mark there is no such information available (comes from xsd:anyURI definition as it allows empty but otherwise it should be URI or URL)
As this not mandatory in SPDX Lite 3.0.1 there should be thought should this be added only when available but now it's always for sake of completeness


kaniini:
This one needs a rebase but looks good

kaniini:
Damn you github!

PR #486, opened 2026-04-04
Replace ``wixl`` by ``WiX``

### LINKED ISSUE #406: Improve windows support

moi15moi:
Fix: https://github.com/pkgconf/pkgconf/issues/406#issuecomment-4187551065

The goal of this PR is to properly build msi on arm.
Since ``wixl`` doesn't support arm (https://gitlab.gnome.org/GNOME/msitools/-/issues/61), I thought we could try to replace it with  ``WiX``.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,14 +43,29 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install meson ninja
 
+      - name: Install WiX
+        run: |
+          dotnet tool install --global wix
+          $wixVersion = (wix --version).Trim()
+          wix extension add -g "WixToolset.UI.wixext/$wixVersion"
+
       - name: Build
         run: |
           meson _build --vsenv
-          meson compile -C _build
+          meson compile -C _build msi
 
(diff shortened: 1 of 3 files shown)


moi15moi:
We should maybe set `_WIN32_WINNT` to properly specify which Windows version we support.

moi15moi:
@kaniini What do you think about my last comment?

moi15moi:
@kaniini What do you think about my last comment?

PR #487, opened 2026-04-11
Support WiX version 7.0.0

moi15moi:
Wow, I can't believe in just 5 days since this PR was merged (https://github.com/pkgconf/pkgconf/pull/486), it is already broken.
WiX 7.0.0 require us to pass the option "-acceptEula". See: https://docs.firegiant.com/wix/osmf/#accept-the-eula-wix-v7-and-later
It does sucks, but there isn't really any good alternative to WiX, so I guess we can live with it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
         run: |
           dotnet tool install --global wix
           $wixVersion = (wix --version).Trim()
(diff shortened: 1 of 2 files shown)


kaniini:
I took a simpler approach in 465175d

moi15moi:
It work on ci, but not when building locally. Thats up to you.

PR #492, opened 2026-04-30
Check buffer API for failure and return errors on failure

Elizafox:
This PR revamps the libpkgconf API to return failures from underlying functions. `pkgconf_buffer_*` functions are checked for failure throughout and many API functions now return `bool` to accommodate this.

This is a follow-up from #489.

Other changes made:
- The CLI using these APIs now more aggressively checks for failures and reports to the user.
- Several docstrings have been fixed, as well as some use-after-free bugs.
- Cleanup of handling of OOM conditions.
- Several potential use-after-free bugs fixed.


Elizafox:
I will split this PR instead into multiple pieces. Closing for now.

PR #495, opened 2026-05-04
New file output wrappers and conversion

Elizafox:
This adds a new set of functions: `pkgconf_output_file_vfmt` and `pkgconf_output_file_fmt`, which act as simple wrappers around `fprintf` and convert the return type to `bool`.

This PR also converts most instances of `fprintf` and adds checking of the output of these functions, especially ones outputting to file descriptors besides `stderr`. If `stderr` fails, there is little we can do, so this is not generally checked. But it is useful to check other FDs, to display a diagnostic to the end user if a write fails (broken pipe, filesystem full, filesystem failure, etc.).

This is not perfect, and things are not always immediately aborted on I/O failure. This will require deeper API changes like in #492, but this is still an improvement.

### DIFF
--- a/cli/bomtool/main.c
+++ b/cli/bomtool/main.c
@@ -35,6 +35,22 @@ static int maximum_traverse_depth = 2000;
 static FILE *error_msgout = NULL;
 static FILE *sbom_out = NULL;
 
+#define OUTPUT_OR_RET(client, f, fmt, ...) \
+    do { \
+        if (!pkgconf_output_file_fmt((f), (fmt), ##__VA_ARGS__)) { \
+            pkgconf_error((client), "bomtool: Could not output to file: %s", strerror(errno)); \
+            return; \
+        } \
+    } while (0)
+
+#define OUTPUT_OR_RET_FALSE(client, f, fmt, ...) \
+    do { \
+        if (!pkgconf_output_file_fmt((f), (fmt), ##__VA_ARGS__)) { \
+            pkgconf_error((client), "bomtool: Could not output to file: %s", strerror(errno)); \
+            return false; \
+        } \
+    } while (0)
+
 static const char *
 environ_lookup_handler(const pkgconf_client_t *client, const char *key)
 {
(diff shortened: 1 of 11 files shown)


cli/bomtool/main.c
```diff
@@ -87,22 +103,26 @@ sbom_name(pkgconf_pkg_t *world)
...
 
-	fprintf(sbom_out, "SPDXVersion: %s\n", spdx_version);
-	fprintf(sbom_out, "DataLicense: %s\n", bom_license);
-	fprintf(sbom_out, "SPDXID: %s\n", document_ref);
-	fprintf(sbom_out, "DocumentName: %s\n", docname);
-	fprintf(sbom_out, "DocumentNamespace: https://spdx.org/spdxdocs/bomtool-%s\n", PACKAGE_VERSION);
-	fprintf(sbom_out, "Creator: Tool: bomtool %s\n", PACKAGE_VERSION);
-
-	fprintf(sbom_out, "\n\n");
+	bool ret = pkgconf_output_file_fmt(sbom_out, "SPDXVersion: %s\n", spdx_version)
```
kaniini: why would we not just use OUTPUT_OR_RET_FALSE here?

libpkgconf/buffer.c
```diff
@@ -36,11 +36,12 @@ target_allocation_size(size_t target_size)
 static void
 buffer_debug(pkgconf_buffer_t *buffer)
 {
-	for (char *c = buffer->base; c <= buffer->end; c++) {
-		fprintf(stderr, "%02x ", (unsigned char) *c);
+	for (char *c = buffer->base; c <= buffer->end; c++)
+	{
+		pkgconf_output_file_fmt(stderr, "%02x ", (unsigned char) *c);
```
kaniini: This does not compile, i would prefer to just leave it as it is as it's a debugging aid for me.

cli/bomtool/main.c
Elizafox: Look below it:

```
	int errno_save = errno;
	free(docname);

	if (!ret)
		pkgconf_error(client, "bomtool: Could not output to file: %s", strerror(errno_save));
	return ret;
```

We need to free the docname if any of that fails regardless.

libpkgconf/buffer.c
Elizafox: Fair enough.

cli/bomtool/main.c
kaniini: yes, i see that.  i am saying to restructure this.

Elizafox: Fixed per discussion elsewhere.

PR #506, opened 2026-05-27
autogen.sh: use /bin/echo for -n

khorben:
On macOS, `echo` from `/bin/sh`'s shell builtin implementation does not support the `-n` flag. This uses `/bin/echo` instead when using `-n`.

Definitely not a big deal, feel free to reject if deemed too risky in terms of portability to rely on the presence of `/bin/echo` like this.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -59,7 +59,7 @@ run_or_die ()
    OPTIONS="$@"
    
    # print a message
-   echo -n "*info* running $COMMAND"
+   /bin/echo -n "*info* running $COMMAND"
    if test -n "$OPTIONS" ; then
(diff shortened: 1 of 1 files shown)


kaniini:
let's switch to printf instead?

autogen.sh
```diff
@@ -59,7 +59,7 @@ run_or_die ()
    OPTIONS="$@"
    
    # print a message
-   echo -n "*info* running $COMMAND"
+   /bin/echo -n "*info* running $COMMAND"
```
kaniini, line 62:
```suggestion
   printf "*info* running %s\n" "$COMMAND"
```

PR #511, opened 2026-05-28
Bump pkgconf version to 3.0.0 as that will be the next release

Elizafox:
Does what it says on the tin. This matches the NEWS file as well.


kaniini:
We should probably cut some release candidates first, lets go with 2.9.50 for now?

PR #517, opened 2026-05-30
Run coverage in the CI and upload it to codecov

moi15moi:
I think it would be beneficial to use meson [coverage](https://mesonbuild.com/Unit-tests.html#coverage) feature to track which part of pkgconf is covered by tests.
Knowing the coverage help to know what tests we should add.

In brief, this PR use meson coverage feature to generate a report that is uploaded to https://app.codecov.io/.
This is how the report looks: https://app.codecov.io/github/pkgconf/pkgconf/tree/moi15moi:Add-coverage

If you are interested in this PR, note that you will need to properly config https://app.codecov.io and add the ``CODECOV_TOKEN`` to the secret (note that you can not require to use a token like written here https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token but it would main anyone could upload a coverage to codecov, so it's a pretty bad idea)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -306,16 +306,27 @@ jobs:
       - name: Update system and add dependencies
         run: |
           apk update
-          apk add build-base meson
+          apk add build-base meson gcovr bash git curl gpg
 
       - name: Build
         run: |
-          meson _build -Dwerror=true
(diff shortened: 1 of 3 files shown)


kaniini:
Seems alright but needs DCO signoff.

PR #520, opened 2026-06-04
manpage: Document when extensions were added to pkgconf

### LINKED ISSUE #514: [man] Document which pkgconf version introduced a field
Currently, https://github.com/pkgconf/pkgconf/blob/ffc80e0364181638d33c9adc0edf3f30054ad036/man/pc.5 document all the possible field, but it doesn't specify in which version those fields where introduced.
I believe it might be a good idea to specify it.

For example, for License.file, Requires.shared, Cflags.shared, Libs.shared, we could specify those where introduced in the version 3.0.0

Elizafox:
Simple documentation update. Information gathered from `NEWS` file and from commit/tag history.

Fixes #514

### DIFF
--- a/man/pc.5
+++ b/man/pc.5
@@ -98,13 +98,13 @@ These flags are always used, regardless of whether static compilation is request
 (optional; fragment list)
 .It Cflags.private
 Required compiler flags for static compilation only.
-(optional; fragment list; pkgconf extension)
+(optional; fragment list; pkgconf extension; introduced in pkgconf 0.9.3)
 .It Cflags.shared
 Required compiler flags for shared compilation only.
-(optional; fragment list; pkgconf extension)
+(optional; fragment list; pkgconf extension; introduced in pkgconf 3.0.0)
 .It Copyright
 A copyright attestation statement.
-(optional; literal; pkgconf extension)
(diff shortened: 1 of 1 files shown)


man/pc.5
```diff
@@ -115,7 +115,7 @@ Required linking flags for this package that are only required when linking
 statically.
 Libraries this package depends on for linking against it statically, which are
 not described as dependencies should be specified here.
-(optional; fragment list)
+(optional; fragment list; introduced in pkgconf 3.0.0)
```
moi15moi: ``Libs.private`` isn't a pkgconf extension and it has been supported since the version 0.7.0.
Did you mean ``Libs.shared``?

kaniini: Yeah, Libs.shared hasn't been documented yet.  We need to fix that and mark it as 3.0.0.

We should also note that you can have a pc(5) file declare its requirements on pkgconf directly for these extensions such as

```
Requires: pkgconf >= 3.0.0, ...
Libs.shared: -lwhatever
```

Elizafox: Whoops! I missed that. I fixed it and added libs.shared. Should probably document libs.shared more thoroughly as @kaniini described though.

kaniini (approved):

PR #525, opened 2026-06-07
CI: Run tests on cygwin

moi15moi:
I believe it is a good idea to run the tests on cygwin because currently, the tests doesn't pass.
I fixed the problem with ``multiline-folding-cflags.test``, but I couldn't found the issue with the ``basic-with-colon.test`` test.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 tests/lib1/dos-lineendings.pc eol=crlf
+tests/lib-sbom-files/*.json eol=lf
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,6 +153,33 @@ jobs:
         run: |
           meson test -v -C _build
 
+  cygwin:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout code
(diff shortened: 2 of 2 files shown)


kaniini:
It's failing because the test runner is truncating the final newline on cygwin.  Not sure why.

moi15moi:
> It's failing because the test runner is truncating the final newline on cygwin. Not sure why.

It truncate the final new line because of: https://github.com/pkgconf/pkgconf/blob/19764053c7c6645ac681d059d055481731e67ef9/tests/test-runner.c#L1369-L1370 (so, instead of removing the \r\n, it just remove the \n).

Even if I remove this line, the test still fail because the expected file contains \r\n, but the test output \n (which is right).
So, in brief, we should simply ignore the difference with line ending for the tests.

kaniini:
Lets rebase this and get it in.

kaniini:
On `feature/cygwin`, I just added a `.gitattributes` directive to specify the line-endings as LF instead of CRLF.  What do you think?

moi15moi:
Sure, it also work.
I will reset my branch on yours.

PR #526, opened 2026-06-08
Use utf-8 codepage on Windows

### LINKED ISSUE #518: [Windows] If PKG_CONFIG_PATH contain a non-ainsi character, pkgconf can't find .pc file
If PKG_CONFIG_PATH contain a non-ainsi character, pkgconf can't find .pc file.
For example,  if I set PKG_CONFIG_PATH to ``C:\Users\moi15moi\Documents\GitHub\vcpkg_😋ffmpeg\installed\x64-windows-static\lib\pkgconfig``, pkgconf won't be able to detect the .pc file in this folder.

moi15moi:
Since [Windows Version 1903 (May 2019 Update)](https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page), it is possible to specify, in a [manifest](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests), the code page of a executable.

I think the best fix would be to replace all the ainsi function (ex: ``fopen`` with ``_wfopen``, ``getenv`` with ``_wgetenv``), but [the maintainer(s)](https://github.com/pkgconf/pkgconf/issues/518#issuecomment-4598233145) seems reticent about that.

So, with this approach, the fix is much smaller then to refactor pkgconf, BUT it's not as perfect as refactoring the codebase because this fix will only work on Windows Version 1903 (May 2019 Update) or more.
Also, when other application will use libpkgconf, they wont have this fix because it is only applied to executable, not dll.

This allow to fix https://github.com/pkgconf/pkgconf/issues/518


Elizafox:
Should we call [`SetConsoleCP`](https://learn.microsoft.com/en-us/windows/console/setconsolecp) / [`SetConsoleOutputCP`](https://learn.microsoft.com/en-us/windows/console/setconsoleoutputcp) as well?

PR #533, opened 2026-06-13
libpkgconf: fragment: don't split UTF-8 sequences when quoting fragments

moi15moi:
fragment_quote() escapes every byte in the range [0x7f, 0xff] with a backslash. For a fragment containing a UTF-8 multibyte sequence (ex: "😋" which is \xF0\x9F\x98\x8B in UTF-8), this inserts a backslash before each byte of the sequence individually (ex: \\\xf0\\\x9f\\\x98\\\x8b for "😋"), producing mangled output instead of the original UTF-8 text.

To detect if the text if UTF-8, let's simply check the LC_CTYPE codeset on POSIX and the ANSI code page on Windows. When it does, narrow the escaped range to DEL (0x7f) only, so bytes 0x80-0xff pass through untouched and multibyte sequences stay intact.

Note: This is not compatible with upstream pkg-config's strdup_escape_shell() (https://gitlab.freedesktop.org/pkg-config/pkg-config/-/blob/d97db4fae4c1cd099b506970b285dc2afd818ea2/parse.c#L591-618). As a result, pkgconf and pkg-config now produce different quoted output for non-ASCII fragments in a UTF-8 locale (but they will still produce the same result if the locale is not UTF-8).

This PR try to fix this comment: https://github.com/mesonbuild/meson/issues/15903#issuecomment-4686542184

PS: This PR was partially created with the assistance of an LLM.

### DIFF
--- a/libpkgconf/fragment.c
+++ b/libpkgconf/fragment.c
@@ -18,6 +18,11 @@
 #include <libpkgconf/stdinc.h>
 #include <libpkgconf/libpkgconf.h>
 
+#ifndef _WIN32
+#include <locale.h>
+#include <langinfo.h>
+#endif
+
 /*
  * !doc
  *
@@ -651,6 +656,41 @@ pkgconf_fragment_filter(const pkgconf_client_t *client, pkgconf_list_t *dest, pk
 	}
 }
 
+/*
+ * !doc
+ *
+ * .. c:function:: bool pkgconf_is_locale_utf8(void)
+ *
+ *    Check whether text is expected to be UTF-8 encoded in the current environment.
+ *
+ *    :return: :code:`true` if text is expected to be UTF-8 encoded, :code:`false` otherwise.
+ */
+bool
+pkgconf_is_locale_utf8(void)
+{
+#ifdef _WIN32
+	return GetACP() == CP_UTF8;
+#else
+	static int cached = -1;
+
+	if (cached < 0)
+	{
+		locale_t loc = newlocale(LC_CTYPE_MASK, "", (locale_t)0);
+
+		if (loc != (locale_t)0)
+		{
+			const char *codeset = nl_langinfo_l(CODESET, loc);
+			cached = codeset != NULL && (!strcasecmp(codeset, "UTF-8") || !strcasecmp(codeset, "UTF8"));
+			freelocale(loc);
+		}
+		else
+			cached = 0;
+	}
+
+	return cached;
+#endif
+}
+
 static void
 fragment_quote(pkgconf_buffer_t *out, const pkgconf_fragment_t *frag)
 {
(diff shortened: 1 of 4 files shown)


t/basic/utf8-pkg-config-path.test
```diff
@@ -1,4 +1,4 @@
 Environment: PKG_CONFIG_PATH=%TEST_FIXTURES_DIR%/lib1/të😋st/lib/pkgconfig
 WantedFlags: cflags define-prefix
 Query: utf8
-ExpectedStdout: -I%TEST_FIXTURES_DIR%/lib1/t\�\�\�\�\�\�st/include
+ExpectedStdout: -I%TEST_FIXTURES_DIR%/lib1/të😋st/include
```
moi15moi, line 4: Note that this test will fail on non UTF-8 locale.

libpkgconf/fragment.c
```diff
@@ -632,14 +637,49 @@ pkgconf_fragment_filter(const pkgconf_client_t *client, pkgconf_list_t *dest, pk
...
+	static int cached = -1;
+
+	if (cached < 0)
+	{
+		locale_t loc = newlocale(LC_CTYPE_MASK, "", (locale_t)0);
+
+		if (loc != (locale_t)0)
+		{
+			const char *codeset = nl_langinfo_l(CODESET, loc);
+			cached = codeset != NULL && (!strcasecmp(codeset, "UTF-8") || !strcasecmp(codeset, "UTF8"));
```
moi15moi, line 683: To detect if a locale is indeed UTF-8, this is what claude suggested me.
The glibc behaviour is to always return ``UTF-8`` (see Application notes here: https://man7.org/linux/man-pages/man7/utf-8.7.html), but claude highly recommended me to not care about the case and the hyphen.

libpkgconf/fragment.c
```diff
@@ -632,14 +637,49 @@ pkgconf_fragment_filter(const pkgconf_client_t *client, pkgconf_list_t *dest, pk
...
+
 static void
 fragment_quote(pkgconf_buffer_t *out, const pkgconf_fragment_t *frag)
 {
 	if (frag->data == NULL)
 		return;
 
 	const pkgconf_buffer_t *src = PKGCONF_BUFFER_FROM_STR(frag->data);
-	const pkgconf_span_t quote_spans[] = {
+	pkgconf_span_t quote_spans[] = {
```
kaniini: Can we find a solution where we keep the quote_spans const?

libpkgconf/fragment.c
```diff
@@ -652,6 +692,11 @@ fragment_quote(pkgconf_buffer_t *out, const pkgconf_fragment_t *frag)
 		{ 0x7f, 0xff },
 	};
 
+	/* If the local is UTF-8 we must not split character over 0x7f because it would add "\" between each bytes.
```
kaniini, line 714: I think a better solution would be to have two different quote_spans, and select the correct one based on whether a UTF-8 locale is presently configured.

t/basic/utf8-pkg-config-path.test
kaniini: We can add some logic to skip running the test if Unicode locale is not configured.

kaniini:
To what extent did you use Claude to generate the PR?  Research or did it generate the patches?

t/basic/utf8-pkg-config-path.test
moi15moi: If you want that, I believe I should rename ``is_locale_utf8`` to ``pkgconf_is_locale_utf8`` and make it as a public api.
I could add a variable ``RequireUtf8Locale`` to [test_keyword_pairs](https://github.com/pkgconf/pkgconf/blob/63115f705432f4b759a537db866743082248370b/tests/test-runner.c#L533-L566) and it would call ``pkgconf_is_locale_utf8``.
Are you ok with that?

moi15moi:
I used claude to know how to detect if the locale is UTF-8 on unix.

kaniini:
Sounds good.  I think it is ok to use it for that kind of research.  My concern is simply making sure that we humans remember how this shit works 😊

t/basic/utf8-pkg-config-path.test
kaniini: Yes, sounds good.

moi15moi: I applied the fix.
Tell me what you think about it.

libpkgconf/fragment.c
moi15moi: I applied the fix.
Tell me what you think about it.

libpkgconf/fragment.c
moi15moi: I applied the fix.
Tell me what you think about it.

kaniini: This is fine for now.  I might do some cleanups this evening.

t/basic/utf8-pkg-config-path.test
kaniini: looks fine.

PR #539, opened 2026-06-21
ci: Build/test the project on macOS/NetBSD/FreeBSD/Haiku/Omnios

moi15moi:
The compilation on macOS was broken in this PR (https://github.com/pkgconf/pkgconf/pull/533).
I was getting this error:
```c
../libpkgconf/fragment.c:682:26: error: call to undeclared function 'nl_langinfo_l'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
  682 |                         const char *codeset = nl_langinfo_l(CODESET, loc);
      |                                               ^
../libpkgconf/fragment.c:682:26: note: did you mean 'nl_langinfo'?
/Applications/Xcode_16.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/_langinfo.h:116:20: note: 'nl_langinfo' declared here
  116 | char    *_LIBC_CSTR   nl_langinfo(nl_item);
      |                       ^
../libpkgconf/fragment.c:682:16: error: incompatible integer to pointer conversion initializing 'const char *' with an expression of type 'int' [-Wint-conversion]
  682 |                         const char *codeset = nl_langinfo_l(CODESET, loc);
      |                                     ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

For whatever reason, we need to include ``xlocale.h`` to be able to call ``nl_langinfo_l``.
See: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/nl_langinfo.3.html

Since there isn't any CI that use clang on a unix platform, let's use macOS to verify that we properly support clang.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,6 +153,79 @@ jobs:
         run: |
           meson test -v -C _build
 
+  QEMU_VMs:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: cpa.sh {0}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os_image: netbsd
+            version: '10.1'
+
+          - os_image: freebsd
+            version: '15.1'
+
+          - os_image: openbsd
+            version: '7.9'
+
+          - os_image: haiku
+            version: 'r1beta5'
+
+          - os_image: omnios
+            version: 'r151056'
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Start VM on ${{ matrix.os_image }}
+        uses: cross-platform-actions/action@v1.3.0
+        with:
+          operating_system: ${{ matrix.os_image }}
+          version: ${{ matrix.version }}
+
+      - name: Install meson
+        run: |
+          case "${{ matrix.os_image }}" in
+            netbsd)
+              sudo pkgin -y install meson
+              ;;
+            freebsd)
+              # Using system compiler
+              sudo pkg -o ASSUME_ALWAYS_YES=yes install meson
+              ;;
+            openbsd)
+              sudo pkg_add meson
+              ;;
+            haiku)
+              pkgman install -y meson
+              ;;
+            omnios)
+              sudo pkg install --deny-new-be --no-backup-be \
+                pkg:/developer/gcc14 \
+                pkg:/library/python-3/meson-313 \
+                pkg:/ooce/developer/ninja
+              ;;
+          esac
+
+      - name: Build
+        run: |
+          if [[ "${{ matrix.os_image }}" == "omnios" ]]; then
+            export PATH=/usr/lib/python3.13/bin:$PATH
+          fi
+          meson _build -Dwerror=true
+          meson compile -C _build
+
+      - name: Run tests
+        run: |
+          if [[ "${{ matrix.os_image }}" == "omnios" ]]; then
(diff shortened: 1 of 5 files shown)


moi15moi:
I put the PR to a draft because I realised that ``nl_langinfo_l`` isn't available on NetBSD and OpenBSD (and some other platform https://www.gnu.org/software/gnulib/manual/html_node/nl_005flanginfo_005fl.html)

To fix this, I fallback to ``nl_langinfo`` which seems to have a better availability (https://www.gnu.org/software/gnulib/manual/html_node/nl_005flanginfo.html).

But, I realised that the compilation wasn't even working on NetBSD, so I also fixed it.

Now, I am trying to understand why the test fail on OpenBSD.

PS: Note that I am not familiar with BSD systems, so please review my fix carefully.

moi15moi:
I decided to open a issue for the issue with OpenBSD: https://github.com/pkgconf/pkgconf/issues/540
I added a CI for NetBSD, FreeBSD, Haiku and Omnios to test as much different toolchain.

kaniini:
I think the _NETBSD_SOURCE thing is wrong.  it is a similar code smell to _WIN32.  we have to put up with Win32, but I expect POSIX conformance out of NetBSD.

moi15moi:
I reverted the commit just so you can see the warning we get.
It's kinda surprising that it doesn't segfault, but it's definitely not normal.

kaniini:
Try rebasing on latest HEAD.  It should fix the NetBSD warnings.

moi15moi:
Good call, https://github.com/pkgconf/pkgconf/commit/9447d89cd9d74f04232138d5fa634a0ecfbdba05 did indeed fix the problem.
By curiosity, how did you know that ``_XOPEN_SOURCE=700`` would solve this? It doesn't seems documented here: https://man.netbsd.org/realpath.3

Do you want the CI to have ``-Dwerror=true``?

kaniini:
Both symbols are defined by SUSv4, which is gated at _XOPEN_SOURCE>=500.

kaniini:
latest HEAD should fix openbsd, also we should block on werror=true imo :)

.github/workflows/test.yml
```diff
@@ -133,6 +133,94 @@ jobs:
...
+            version: 'r1beta5'
+
+          - os_image: omnios
+            version: 'r151056'
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Start VM on ${{ matrix.os.name }}
```
kaniini:
```suggestion
      - name: Start VM on ${{ matrix.os_image }}
```

moi15moi: Good catch

kaniini:
It seems like pkg_add hangs on OpenBSD.  Maybe we can choose a more reliable mirror?

.github/workflows/test.yml
```diff
@@ -133,6 +133,99 @@ jobs:
...
+          case "${{ matrix.os_image }}" in
+            netbsd)
+              sudo pkgin -y install meson
+              ;;
+            freebsd)
+              # Using system compiler
+              sudo pkg -o ASSUME_ALWAYS_YES=yes install meson
+              ;;
+            openbsd)
+              sudo pkg_add meson
```
kaniini:
```suggestion
              sudo pkg_add https://cdn.openbsd.org/%m/meson
```

PR #550, opened 2026-06-25
api-tests: add personality API unit tests

Elizafox:
Add coverage for pkgconf_cross_personality_default, deinit (null, refcount, non-default free path), and find (direct file path, XDG data dir lookup, invalid triplet, missing triplet fallback to default).

Covers all 7 keyword types parsed by the personality file parser.


Elizafox:
Needs more work on Windows before I land this.

EDIT: done.

PR #551, opened 2026-06-25
spdxtool: introduce OOM testing

Elizafox:
This adds an spdxtool out-of-memory fault-injection test.

## Memory allocator fault injector
This adds a reusable allocation-failure injection harness (tests/api/oom.h). It wraps malloc/calloc/realloc/strdup/strndup via the linker's `--wrap` so a test can fail the N-th allocation a target makes, plus an exhaustive driver that injects a failure at every successive allocation and asserts the target reports the error gracefully (and, under ASAN, leaks nothing on the way out).

tests/api/test-oom.c applies it to spdxtool's constructors, the serializer value model, and the self-contained `*_to_object` functions.  It is built only on glibc Linux: it needs a linker that supports `--wrap` (not macOS ld64, BSD's, or MSVC) and reliable LeakSanitizer (not musl/Alpine), so it runs on the glibc sanitizer CI job and is skipped elsewhere.

These tests are omitted from autotools as autotools for pkgconf are deprecated and codecov is not run from autotools.

This lifts spdxtool allocation-failure coverage substantially (serialize.h to 100%, core.c to 85%, simplelicensing.c to 97%, util.c to 95%).

## Fixes for spdxtool
Several spdxtool error paths leaked memory when an allocation failed:

- `serialize value_object()` / `value_array()` returned `NULL` on `calloc()` failure without freeing the container they document as "taking ownership of". Every `*_to_object()` caller does `ret = value_object(list); list = NULL;` unconditionally, so a failure here orphaned the entire built object/array. Free the container on failure so ownership transfer is truly unconditional.

- `object_add_take()` freed only the *contents* of its partially-built object entry (via `object_free()`, which by contract does not free the struct itself) when the key `strdup()` failed, leaking the entry struct. Free it.

- `creation_info_to_object()` orphaned its `createdBy` array when an earlier `object_add_string()` short-circuited the `&&` chain before the array was handed to the object list. Split the chain and free it on that path.

All found by the new spdxtool OOM fault-injection test under AddressSanitizer.

### DIFF
--- a/cli/spdxtool/core.c
+++ b/cli/spdxtool/core.c
@@ -222,8 +222,15 @@ spdxtool_core_creation_info_to_object(pkgconf_client_t *client, const spdxtool_c
 
 	if (!(spdxtool_serialize_object_add_string(object_list, "type", creation->type) &&
 		spdxtool_serialize_object_add_string(object_list, "@id", creation->id) &&
-		spdxtool_serialize_object_add_string(object_list, "created", creation->created) &&
-		spdxtool_serialize_object_add_array(object_list, "createdBy", created_by) &&
+		spdxtool_serialize_object_add_string(object_list, "created", creation->created)))
+	{
+		/* created_by has not been handed to the object list yet */
+		spdxtool_serialize_array_free(created_by);
+		goto err;
+	}
+
+	/* object_add_array takes ownership of created_by, freeing it on failure */
+	if (!(spdxtool_serialize_object_add_array(object_list, "createdBy", created_by) &&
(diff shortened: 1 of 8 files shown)


kaniini:
The fuzzer already has a similar OOM injection path.  Can we put this in a common implementation?

Elizafox:
> The fuzzer already has a similar OOM injection path. Can we put this in a common implementation?

I can investigate that.

Elizafox:
@kaniini what do you think of this iteration? I have rewritten the OOM tests to use the fuzzer and augmented it a bit.

PR #552, opened 2026-06-25
codecov: exempt API tests from coverage

Elizafox:
These are test programs that should not count towards overall coverage.

No functional changes.


moi15moi:
I recommand you to read: https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage

kaniini:
Yes, I think we want to include tests in the coverage.  The reason I exempted the runner was because it was noise that we can drill down on later.

PR #554, opened 2026-06-26
spdxtool: Add /Core/Tool

illuusio:
This one is definitely too late for 3.0 and this PR is more to make `spdxtool` feature wise with `bomtool`. Also update tests to correlate the change.

### DIFF
--- a/cli/spdxtool/core.c
+++ b/cli/spdxtool/core.c
@@ -127,6 +127,116 @@ err:
 /*
  * !doc
  *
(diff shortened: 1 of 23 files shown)


illuusio:
I need to investigate about coverage and why there is still memory leaks as there shouldn't be.

illuusio:
OmniOS seems to `HTTP 502` so there something on their side. Otherwise it seems that I nailed down leaks.

PR #564, opened 2026-07-12
build: distribute `pkgconf.manifest.in` and `pkgconf.rc.in`

bgilbert:
Meson builds need these on Windows and they weren't being included in the dist tarball.


kaniini:
I will accept this and backport it to release/3.0, but 3.1 won't use autotools-generated tarballs.

PR #565, opened 2026-07-13
cmake: ship a find_package(pkgconf CONFIG) package config

Elizafox:
Add `cmake`/`pkgconf-config.cmake(.in)`, installed by meson to `$libdir/cmake/pkgconf`, so projects can locate pkgconf with `find_package(pkgconf CONFIG)` and query modules through it.

This CMake module provides `pkgconf_check_modules()` and `pkgconf_search_module()`, modelled on CMake's ownFindPkgConfig.cmake. However, it always uses pkgconf instead of whatever pkg-config is first on $PATH. This also provides thin wrappers over the single-module pkgconf options (`--variable``, `--print-variables`, `--path`, version checks, `--print-provides`, `--print-requires`, `--link-abi`, `--list-all`). `IMPORTED_TARGET` builds a `PkgConf::<prefix>` interface target from the resulting flags.

A cmake-macros CI job builds and installs pkgconf, then exercises the macros against pkgconf's own `libpkgconf.pc` and a real third-party package (zlib), and checks that a `REQUIRED` nonexistent module fails configure. The CI job runs on Linux (static and shared) and macOS.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,9 +102,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include: [
-          { msystem: MINGW64, arch: x86_64},
-        ]
+        include: [{ msystem: MINGW64, arch: x86_64 }]
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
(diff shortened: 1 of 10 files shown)


Elizafox:
I will resubmit this as a more comprehensive PR in a moment.

PR #567, opened 2026-07-14
cmake: add pkgconf-native package discovery, querying, and .pc generation

Elizafox:
CMake already ships `FindPkgConfig`, but it is intentionally implementation-neutral: it may find `pkgconf`, FDo `pkg-config`, `u-config`, or another compatible executable. That remains appropriate for projects that only need generic pkg-config functionality.

This PR adds pkgconf-owned CMake integration for projects that specifically want or need pkgconf behavior, enhanced querying, or need to produce .pc files from CMake targets.

This CMake package provides:

- `pkgconf::pkgconf`, an imported executable target for the selected binary.
- `pkgconf_check_modules()` and `pkgconf_search_module()` for consuming .pc dependencies.
- Per-query `pkgconf` controls such as `STATIC`, `SHARED`, `ENV_ONLY`,` NO_UNINSTALLED`, `NO_PROVIDES`, `IGNORE_CONFLICTS`, variable overrides, additional search paths, and traversal limits.
- Queries for variables, package paths, versions, `Provides`, public/private Requires, link ABI, and installed package enumeration.

Projects can also generate and install .pc files from CMake targets:

```cmake
pkgconf_generate_pc(testlib
  DESCRIPTION "Test library"
  RELOCATABLE
  INSTALL)
```

When a CMake target links those imported targets, `pkgconf_generate_pc()` translates its public and private link interface into `Requires` and `Requires.private`.

The generator derives metadata from the target and project with explicit overrides. It can do conventional `PREFIX`-based output (useful for system packages) or relocatable output (useful for SDKs, Windows packages, bundles, etc.).

This requires CMake 3.13 or newer.


dg0yt:
I assume this is somewhere marked as experimental and subject to change?

CMake has two broken or imcomplete pc file interfaces at the moment. No support for CMake multi-config mode. And projects which generate pc files often poorly deal with generator expressions.

PR #573, opened 2026-07-18
Add OS/2 support

komh:
Hi/2.

These are the patches for OS/2 support.

Review, please...

### DIFF
--- a/cli/main.c
+++ b/cli/main.c
@@ -82,7 +82,7 @@ deduce_personality(char *argv[])
 	if (i != NULL)
 		argv0 = i + 1;
 
-#if defined(_WIN32) || defined(_WIN64)
+#if defined(_WIN32) || defined(_WIN64) || defined(__OS2__)
 	i = strrchr(argv0, '\\');
 	if (i != NULL)
 		argv0 = i + 1;
--- a/cli/main.c
+++ b/cli/main.c
@@ -269,7 +269,7 @@ main(int argc, char *argv[])
 		return EXIT_FAILURE;
 	}
 
-#ifdef _WIN32
+#if defined(_WIN32) || defined(__OS2__)
 	/* When running regression tests in cygwin, and building native
 	 * executable, tests fail unless native executable outputs unix
 	 * line endings.  Come to think of it, this will probably help
--- a/libpkgconf/libpkgconf-api.h
+++ b/libpkgconf/libpkgconf-api.h
@@ -12,6 +12,12 @@
 # else
 #  define PKGCONF_API __declspec(dllimport)
 # endif
+#elif defined(__OS2__)
+# if defined(LIBPKGCONF_EXPORT) || defined(DLL_EXPORT)
+#  define PKGCONF_API __declspec(dllexport)
+# else
+#  define PKGCONF_API
+# endif
 #else
 # define PKGCONF_API __attribute__((visibility("default")))
 #endif
--- a/libpkgconf/pkg.c
+++ b/libpkgconf/pkg.c
@@ -121,7 +121,11 @@ pkg_get_parent_dir(pkgconf_pkg_t *pkg)
 		 *     /bar (absolute)  |  /foo/link (absolute)  |         /bar (absolute)
 		 *   ../bar (relative)  |  /foo/link (absolute)  |  /foo/../bar (relative)
 		 */
-		if ((sourcebuf[0] != '/') && strcmp(targetdir, "."))
+		if ((sourcebuf[0] != '/') &&
+#ifdef __OS2__
+			!(sourcebuf[0] != '\0' && sourcebuf[1] == ':') &&
+#endif
+			strcmp(targetdir, "."))
 		{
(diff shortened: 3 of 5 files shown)


cli/main.c
```diff
@@ -21,6 +21,10 @@
 #include "getopt_long.h"
 #include "core.h"
 
+#ifdef __OS2__
```
kaniini: Please move these sorts of platform-specific abstractions to `libpkgconf/stdinc.h`.

cli/main.c
```diff
@@ -277,6 +281,11 @@ main(int argc, char *argv[])
 	 */
 	_setmode(fileno(stdout), O_BINARY);
 	_setmode(fileno(stderr), O_BINARY);
+#elif __OS2__
```
kaniini: Fold this into the `_WIN32` block.

libpkgconf/pkg.c
```diff
@@ -123,6 +123,18 @@ pkg_get_parent_dir(pkgconf_pkg_t *pkg)
 		 */
 		if ((sourcebuf[0] != '/') && strcmp(targetdir, "."))
 		{
+#ifdef __OS2__
```
kaniini, line 126: This should match Windows behavior, as OS/2 paths and Windows paths are ~same.  If you disagree, please explain.

libpkgconf/stdinc.h
```diff
@@ -101,8 +101,13 @@
 # else
 #  define PKGCONF_ITEM_SIZE (4096 + 1024)
 # endif
-# define PKG_CONFIG_PATH_SEP_S   ":"
-# define PKG_DIR_SEP_S   '/'
+# ifdef __OS2__
```
kaniini: Factor out the Windows definitions into a shared block.  In general, please avoid mixing OS/2 quirks into the POSIX code path.  Quirks should be handled as we have done with the Windows port.

kaniini (changes requested):
Looks mostly OK but there are some areas that need improvement.

kaniini:
Additionally, the commit messages need Signed-off-by trailers.  Please read [CONTRIBUTING.md](https://github.com/pkgconf/pkgconf/blob/main/CONTRIBUTING.md).

cli/main.c
komh: Ok.

cli/main.c
komh: Ok.

libpkgconf/pkg.c
komh: On OS/2, a symbolic link is supported, and it may have a drive name. So, it's necessary to handle a drive letter.

libpkgconf/stdinc.h
komh: Ok.

komh:
I would do so. Thanks!

libpkgconf/pkg.c
kaniini: ok, if it is necessary, fine, but that also seems relevant to modern Windows, so we should fix it for the general case.

also, I do not feel that the way this is written is very clear as to what is going on with the fallthrough.

libpkgconf/libpkgconf-api.h
```diff
@@ -12,6 +12,12 @@
 # else
 #  define PKGCONF_API __declspec(dllimport)
 # endif
+#elif defined(__OS2__)
+# if defined(LIBPKGCONF_EXPORT) || defined(DLL_EXPORT)
+#  define PKGCONF_API __declspec(dllexport)
+# else
+#  define PKGCONF_API
```
kaniini, line 19: OS/2 really does not require dllimport?  Doesn't it use the same PE/COFF format as NT with the same linking design?

libpkgconf/pkg.c
```diff
@@ -476,21 +485,13 @@ is_path_prefix_equal(const char *path1, const char *path2, size_t path2_len)
 {
 #ifdef _WIN32
 	return !_strnicmp(path1, path2, path2_len);
+#elif defined(__OS2__)
```
kaniini: As we define `strncasecmp` for Windows in `stdinc.h`, you can fold the `_WIN32` block into this one.

kaniini (changes requested):
A few more requested changes!

libpkgconf/libpkgconf-api.h
komh: dllimport is not necessary, and OS/2 uses OMF(OBJ) and AOUT format.

libpkgconf/pkg.c
komh: Ok.

libpkgconf/pkg.c
komh: Any suggestions?

kaniini: I'll go ahead and merge & rewrite it.

kaniini (approved):
Rebased onto current main; conflict resolved against the current buffer API. Local Meson build and all 32 tests pass.

PR #585, opened 2026-08-02
Create meson-test.yml

### LINKED ISSUE #581: Run meson and cmake tests as a CI task
As both the Meson and CMake tests evaluate assumptions about pkgconf's behavior, and we want to avoid regressing them, we should run their test suites with a locally built pkgconf.

This allows us to avoid regressions such as #579.

moi15moi:
Run meson test to try to catch some potential issue/regression our tests didn't catch This partially fix https://github.com/pkgconf/pkgconf/issues/581

I opened this PR in draft because meson tests still fail even with the latest fix (https://github.com/pkgconf/pkgconf/commit/48f04446dea901990e1ec4b0b4778c0cfeba3101)

### DIFF
--- a/.github/workflows/meson-test.yml
+++ b/.github/workflows/meson-test.yml
@@ -1,0 +1,61 @@
+name: meson tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  meson-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout pkgconf
+        uses: actions/checkout@v7
+        with:
+          path: pkgconf
+
+      - name: Checkout meson
+        uses: actions/checkout@v7
+        with:
+          repository: mesonbuild/meson
+          path: meson
+
+      - name: Install build tools
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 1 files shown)


moi15moi:
For the moment, i only run the test on ubuntu, but i plan to also run the test on windows when https://github.com/pkgconf/pkgconf/issues/586 will be fixed.

kaniini:
I got it working but I need you to go adjust the commit message for your part, then we can merge it!

moi15moi:
Yes, sorry about that.
I didnt realize that when uninstalling pkg-config, it was also uninstalling glib, so i thought there was still a problem.
I am trying to also run the test in msys2, but for the moment, i have many errors.

kaniini:
Let's go ahead and merge this since it is already working and useful surveillance.  We can do MSYS2 in a followup.

moi15moi:
Ok. Do you want me to squash the commit?

moi15moi:
I found a way to run test on windows (msys2), but it needs this fix: https://github.com/mesonbuild/meson/pull/16073

PR #590, opened 2026-08-18
Restructure the CI to use dist tarball for other builds

### LINKED ISSUE #569: Restructure the CI to use dist tarball for other builds
Right now we just run the CI on the git repo for all jobs, including the dist job. Instead, we should seed the other jobs off the dist tarball. This helps test our packaging and ensures things work as intended on an actual distribution tarball.

Elizafox:
This restructures the CI to use the dist tarball for other builds.

Fixes #569.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   msvc:
+    needs: dist-meson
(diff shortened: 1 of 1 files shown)


Elizafox:
That build failure is spurious. That test should be re-run.

PR #591, opened 2026-08-18
pkgconf.manifest.in: Set longPathAware to true

moi15moi:
Since Windows Version 1607, if we specify longPathAware in the manifest, it remove the limitation of MAX_PATH (when using fopen or any other API that is manipulating file/directory, the path could be maximum 260 characters).

For more information, see: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

### DIFF
--- a/pkgconf.manifest.in
+++ b/pkgconf.manifest.in
@@ -4,6 +4,7 @@
   <application>
     <windowsSettings>
       <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
+      <longPathAware  xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>
     </windowsSettings>
   </application>
 </assembly>


moi15moi:
@Elizafox Do you have an idea why ``cmake-macros`` run on ubuntu suddently fail?
I'm sure it's unrelated to this PR.

Elizafox:
> @Elizafox Do you have an idea why `cmake-macros` run on ubuntu suddently fail? I'm sure it's unrelated to this PR.

@moi15moi try merging from base branch

moi15moi:
The master branch as the same problem, see: https://github.com/pkgconf/pkgconf/actions/runs/32429834608

Elizafox:
I found the problem: the CI runner looks for zlib.pc, but zlib.pc may exist as a Conda package cache file.

I will fix this to locate the path deterministically using the package manager.

Elizafox:
It looks like this is already fixed in main (I already did it and totally forgot lol). It was the result of a non-deterministic location of zlib.pc and finding conda files spuriously. This absolutely should be fixed in main.

@moi15moi try merging now; this should be fixed as of #590

PR #592, opened 2026-08-20
tests: stop shipping a non-ASCII path in the release tarball

mertefesensoy:
### The problem

`tests/lib1/të😋st/lib/pkgconfig/utf8.pc` is the only non-ASCII path in the tree. `git archive` and therefore both `meson dist` and GitHub's generated release tarballs stores that name as raw UTF-8 in the 100-byte ustar `name` field. The
full member name stays under 100 bytes (56 with a `pkgconf-3.0.5/` prefix, 64 with GitHub's `pkgconf-pkgconf-<sha>/` prefix), so no pax `path=` record is emitted, and git never emits an `hdrcharset` attribute:

```
typeflag '5' magic ustar  name b'pkgconf-3.0.5/tests/lib1/t\xc3\xab\xf0\x9f\x98\x8bst/'
typeflag '5' magic ustar  name b'pkgconf-3.0.5/tests/lib1/t\xc3\xab\xf0\x9f\x98\x8bst/lib/'
typeflag '5' magic ustar  name b'pkgconf-3.0.5/tests/lib1/t\xc3\xab\xf0\x9f\x98\x8bst/lib/pkgconfig/'
typeflag '0' magic ustar  name b'pkgconf-3.0.5/tests/lib1/t\xc3\xab\xf0\x9f\x98\x8bst/lib/pkgconfig/utf8.pc'
```

Reading such an archive, libarchive falls through to `archive_string_default_conversion_for_read()` and transcodes the header name from the OEM code page to the ANSI code page. On a Windows host whose ANSI and OEM code pages differ in practice, a non-UTF-8 system locale the conversion fails, the wide pathname is left unset, and `cleanup_pathname()` in `archive_write_disk_windows.c` rejects it with `Invalid empty pathname`, aborting the whole extraction rather than just that member.

### Scope - please read before weighing this

This is a libarchive bug, not a pkgconf bug, and **the relevant part of it is already fixed upstream**. Kitware commit `f590010`, "libarchive: Restore tolerance of archives with non-OEM paths on Windows" (24 Jan 2026), fixes exactly this path; it shipped in CMake 3.31.11, 4.0.6, 4.1.5 and 4.2.3. Anything extracting through a CMake at or above those versions which includes vcpkg, since it downloads its own CMake when the system one is too old is unaffected.

The fix was upstreamed, not just vendored: the same author landed it in libarchive itself on 6 Feb 2026 ("archive_string_append_from_wcs: Fix silently lossy conversions on Windows" and two companions), with further hardening on 6 Aug 2026. So Windows' own bundled `tar.exe` picks it up whenever Microsoft refreshes its libarchive. What remains is propagation lag — a host pinned to an older CMake, or an un-refreshed system `tar.exe` and the workarounds are well known: update CMake, extract with 7-Zip, or enable UTF-8 worldwide language support.

So this PR is not a bug fix; it is a question of whether pkgconf wants its release artifacts to stay ASCII-only as a matter of policy. If the answer is no, please close it no argument from me.

### Affected releases

Counting non-ASCII entries in `git ls-tree -r` per tag:

| tag | non-ASCII paths |
| --- | --- |
| `pkgconf-2.5.1` and earlier | 0 |
| `pkgconf-2.9.90` | 1 |
| `pkgconf-3.0.5` | 1 |
| `main` (e8a6051) | 1 |

### The change

**1. `tests: materialise the UTF-8 PKG_CONFIG_PATH fixture at test time`**

- The directory name is spelled out as escaped bytes in a new `PKGCONF_TEST_UTF8_DIR`
  in `tests/test-runner.c` (`"t" "\xc3\xab" "\xf0\x9f\x98\x8b" "st"`, written as
  separate string literals so no escape can run into an adjacent hex digit), and
  exposed to `.test` files as a `%UTF8_DIR%` substitution alongside the existing
  `%TEST_FIXTURES_DIR%` / `%DIR_SEP%` / `%PWD%`.
- `t/basic/utf8-pkg-config-path.test` now builds the tree in its per-test temporary
  directory through the existing `SetupMkdir` / `SetupCopy` keywords. The whole file
  is ASCII:

  ```
  SetupMkdir: %UTF8_DIR%/lib/pkgconfig
  SetupCopy: lib-utf8/utf8.pc %UTF8_DIR%/lib/pkgconfig/utf8.pc
  Environment: PKG_CONFIG_PATH=%UTF8_DIR%/lib/pkgconfig
  WantedFlags: cflags define-prefix
  Query: utf8
  ExpectedStdout: -I%UTF8_DIR%/include
  RequireUtf8Locale: true
  ```

- The `.pc` moves to `tests/lib-utf8/utf8.pc` its own fixture directory rather
  than `tests/lib1/`, so it does not become a resolvable package for every test that
  searches `tests/lib1`. Its contents are unchanged.
- `.gitattributes` gains a note recording the constraint.

Because meson, `Makefile.lite` and the cmake tests all drive the same `test-runner`
binary, this one change covers every test path; no build-system changes are needed.

**2. `ci: reject non-ASCII paths in the source tree and the dist tarballs`**

Guards in `dist-meson` and `release-tarballs.yml`. Happy to drop this commit — it is
only worth having if the policy in the first commit is one you want enforced, and it
conflicts textually with #590.

### Verification

- `git archive` on `main`: 4 header names contain a byte > 0x7F. On this branch:
  **0**, out of 818 headers.
- `meson dist` tarball: 817 members, none with a byte > 0x7F.
- Full suite green on the git tree under `LC_ALL=C.utf8` (32/32) and under
  `LC_ALL=POSIX` (32/32, with `utf8-pkg-config-path` reporting `SKIP`).
- Full suite green when unpacked from the `meson dist` tarball and built with
  `-Dwerror=true` (32/32), `utf8-pkg-config-path` reporting `PASS`.
- `make -f Makefile.lite check` (pkgconf-lite): 0 failures,
  `PASS: utf8-pkg-config-path.test`.
- Coverage is unchanged: with the expected output deliberately falsified, the runner
  reports the real result as `stdout: [-Itë😋st/include]` a real non-ASCII
  `PKG_CONFIG_PATH` is still being resolved.

The `RequireUtf8Locale: true` gate is retained, and now also keeps the `mkdir()`
from running on a Windows host whose ANSI code page is not `CP_UTF8`.


moi15moi:
That seems overkill. It's just a tar bug (see https://superuser.com/a/1835267/1729132 + https://github.com/haskell/tar/issues/6) on Windows.
Just use any other tool to extract the tar content (ex: 7zip) and you will be fine.

mertefesensoy:
> That seems overkill. It's just a tar bug (see https://superuser.com/a/1835267/1729132 + [haskell/tar#6](https://github.com/haskell/tar/issues/6)) on Windows. Just use any other tool to extract the tar content (ex: 7zip) and you will be fine.

You're right about the cause, and I agree it isn't pkgconf's bug. A `ustar` name field carries no charset declaration, so on Windows libarchive falls back to `archive_string_default_conversion_for_read()` and decodes header names with the machine's legacy code page. Modern 7-Zip assumes UTF-8 for tar names and handles the archive fine that part of your point is exactly correct.

The problem is that the people hitting this don't get to pick the tool. vcpkg selects its extractor purely by extension: in `vcpkg-tool/src/vcpkg/archives.cpp`, `.gz` maps to `ExtractionType::Tar`, and on Windows that branch unconditionally calls `Tools::TAR` (`tar.exe`). 7-Zip is only reachable for `.zip`, `.7z`, `.exe` and `.nupkg`. There is no flag, setting or triplet variable to change it `find_system_tar()` just takes the first `tar` on `PATH`, so the only lever is shadowing `tar.exe` with a wrapper, which isn't something you can ask every downstream consumer to do.

It also isn't limited to people who want pkgconf itself. Every port that host-depends on it inherits the failure; vcpkg's `sentry-native` port is how I ran into it, and `vcpkg install sentry-native:x64-windows` puts `pkgconf@3.0.3` in the install plan on a Windows target.

mertefesensoy:
The same archive built from `pkgconf-2.5.1` extracts cleanly, and so does 3.0.3 with the four non-ASCII entries removed.

To be clear about the ask I am **not** suggesting you drop the test. The coverage is worth having and `RequireUtf8Locale` already gates it correctly. The narrow request is that the fixture not be shipped *inside the release tarball* create `tests/lib1/të😋st/...` at test time (a meson `custom_target`, or the runner creating the directory and pointing `PKG_CONFIG_PATH` at it), with the directory name spelled as escaped bytes in the source. Same test, same assertion, and `git archive` output stays ASCII-only, so the release keeps extracting everywhere until libarchive changes its default.

A `.gitattributes` `export-ignore` entry would also keep the tarball clean, but on its own it breaks the test for anyone building from the tarball, so it isn't sufficient by itself.

moi15moi:
From what I can see, there is a tracking issue about this on vcpkg repos: https://github.com/microsoft/vcpkg/issues/49338

dg0yt:
- The vcpkg port of pkgconf does not use the official dist tarballs but the on-the-fly tarball from the git tag.
- Extracting tarballs in ports is done via `cmake -E tar`. CMake does indeed use libarchive.

moi15moi:
Does vcpkg download cmake or it simply use the system installed cmake?

If it simply use the system installed cmake, I guess, @mertefesensoy, you can simply update your cmake version to 3.31.11 or above.

dg0yt:
> Does vcpkg download cmake or it simply use the system installed cmake?

It downloads cmake when the system one is too old (and an official build is available at all).

IIUC this bug is for plain Windows. There is no system package manager. OTOH there are installations of CMake and vcpkg managed by the Visual Studio installer. They may lag with regard to the freestanding vcpkg repo.

mertefesensoy:
you are both right, and I have corrected the PR description.

I went and found the fix. Kitware commit `f590010`, "libarchive: Restore tolerance of archives with non-OEM paths on Windows" (bradking, 24 Jan 2026), in `Utilities/cmlibarchive/libarchive/archive_string.c` the exact conversion path
this fails on. It shipped in CMake 3.31.11, 4.0.6, 4.1.5 and 4.2.3.

It did not stop at CMake's vendored copy either. The same author upstreamed the work to libarchive on 6 Feb 2026 "archive_string_append_from_wcs: Fix silently lossy conversions on Windows", "archive_mstring_get_mbs_l: Restore error on failure to convert from WCS/UTF-8", and "archive_string_conversion_charset_name: Fix calls for current locale" with further hardening in "Fix empty wide string conversion on Windows" (6 Aug 2026). So the defect is fixed at the source, not merely routed around in CMake, and Windows' own bundled `tar.exe` picks it up whenever Microsoft refreshes its libarchive.

Combined with @dg0yt's point that vcpkg downloads its own CMake when the system one is too old, that leaves propagation lag rather than a live bug.

mertefesensoy:
- I attributed this to vcpkg without a citation. microsoft/vcpkg#42771, the pkgconf-specific issue, is a different failure with no non-ASCII component. The real tracking issue is the one moi15moi linked, microsoft/vcpkg#49338, and it scopes the trigger more precisely than I did: a non-UTF-8 *system locale*, not merely differing code pages.
- I wrote that pkgconf "cannot be built from the release tarball on that machine at all", flatly and with no version qualifier, seven months after the fix had shipped. That was overstated and I should have checked before writing it.

So I no longer think this is a bug fix, and I would not argue it as one. The only question it still poses is whether pkgconf wants release artifacts to stay ASCII-only as a matter of policy. The cost happens to be near zero here the fixture is generated at test time, the test still resolves a genuinely non-ASCII `PKG_CONFIG_PATH`, and it still skips off a UTF-8 locale but near-zero cost is not by itself a reason to take a patch.

Happy to close this, and happy to drop the CI commit and keep only the fixture change if the policy is worth having but the guards are not. Whichever you prefer please just say, and thank you both for your feedback.

Elizafox:
I encountered this problem in #590

My solution to the problem was to use Python to extract the tarball until such a time that Microsoft ships an updated `tar` implementation that can properly handle UTF-8.

I believe removing a meaningful and useful Unicode test is the wrong solution to this problem. In addition, this is not pkgconf's bug. I vote to close, but update the documentation.

kaniini:
pkgconf has made no commitment to use ASCII-only paths.

Also this PR looks very LLM-assisted.  Have you read the [contributing guide](https://github.com/pkgconf/pkgconf/tree/main/CONTRIBUTING.md)?

As a matter of general policy we do not accept LLM-generated PRs, basically for precisely the reason on display here: the LLM makes assumptions that are not necessarily aligned with the policy set by maintainers (or in this case, lack of policy set by maintainers).

mertefesensoy:
> pkgconf has made no commitment to use ASCII-only paths.
> 
> 
> 
> Also this PR looks very LLM-assisted.  Have you read the [contributing guide](https://github.com/pkgconf/pkgconf/tree/main/CONTRIBUTING.md)?
> 
> 
> 
> As a matter of general policy we do not accept LLM-generated PRs, basically for precisely the reason on display here: the LLM makes assumptions that are not necessarily aligned with the policy set by maintainers (or in this case, lack of policy set by maintainers).

English is not my first language. So using ai to translate what i try to say is convinient for communication.But that is not the case for the code(It's not like a million new line mega pr). This PR aims to fix a problem that i am facing for another project, and not change(or break) anything in pkgconf.

mertefesensoy:
> I encountered this problem in #590
> 
> 
> 
> My solution to the problem was to use Python to extract the tarball until such a time that Microsoft ships an updated `tar` implementation that can properly handle UTF-8.
> 
> 
> 
> I believe removing a meaningful and useful Unicode test is the wrong solution to this problem. In addition, this is not pkgconf's bug. I vote to close, but update the documentation.

I agree with that and this PR does not remove it.

> The narrow request is that the fixture not be shipped inside the release tarball create tests/lib1/të😋st/... at test time (a meson custom_target, or the runner creating the directory and pointing PKG_CONFIG_PATH at it), with the directory name spelled as escaped bytes in the source.

kaniini:
Closing as there is no policy commitment for ASCII paths on pkgconf tarballs.

PR #598, opened 2026-08-30
Avoid space in MSI INSTALLFOLDER

dg0yt:
Probably a debatable improvement:

I try to use the official MSI package for vcpkg.
The package is not installed, but extracted with `msiexec` as "administrative" installation.
This gives a runnable `pkgconf.exe` and `pccritic.exe`, but...
The filepath always contain a space due to the `pkgconf 3.0.x` directory.
When the filepath is passed literally into autotools builds via the `PKG_CONFIG` environment variable, at least some builds will not be able to invoke pkgconf because their scripts expand space into multiple words.

Among various ways to resolve the issue, there is one simple mitigation:
Avoid the space.

### DIFF
--- a/pkgconf.wxs.in
+++ b/pkgconf.wxs.in
@@ -19,7 +19,7 @@
       </Feature>
 
       <StandardDirectory Id="ProgramFiles6432Folder">
-          <Directory Id="INSTALLFOLDER" Name="pkgconf @VERSION@" />
+          <Directory Id="INSTALLFOLDER" Name="pkgconf-@VERSION@" />
       </StandardDirectory>
 
       <ComponentGroup Id="ProductComponents" Directory="INSTALLFOLDER">


moi15moi:
Note that this PR will break tool that are assuming that the INSTALL_FOLDER won't change.
Ex: https://github.com/ScoopInstaller/Main/blob/69590e8d6db2101dc5d10593176c5e80974d5436/bucket/pkgconf.json#L37-L48
But, it isn't too bad.

dg0yt:
Well, the folder will change anyways when the version number changes ;-)

moi15moi:
Read the code i shared. It does set the version number dynamically.

dg0yt:
> Read the code i shared. 

Let's not talk like this. I read it twice before answering, with a sufficient break to avoid jumping into conclusions.

> It does set the version number dynamically.

IMHO that's *internal* convenience.
For the next update, they must update L2-20 (5x `3.0.6`, 3x `hash`).
And every *external* user (which does not rely on redirection through a shim) also needs to move from `3.0.6` to the next identifier.

So that's my point: Getting the (filepath of the) next version doesn't happen by accident.

I don't dispute that there is one-time extra inconvenience.

kaniini:
I have no preference, but isn't "name version" more common on Windows than "name-version"?

dg0yt:
Checking `C:\Program Files` on a random computer ...
(To be fair, not much installed here ...)
There are  
0  `<name>-<version>`  
0  `<name> <version>`  (but I know that the scheme exists)  
k  `<name>`  e.g. `Git`  
n  `<vendor> <name>`  e.g. `Microsoft Visual Studio`  
m  `<CamelCaseName>`  e.g. `WindowsPowerShell`  

Some dirs have the version as a subdir  e.g. `Microsoft Visual Studio\2022`.

Issue #576, opened 2026-07-20
Variables are returned from multiple modules


mmuetzel:
Given the pkg-config files in this archive: [mumps-_to.zip](https://github.com/user-attachments/files/30196327/mumps-_to.zip)

With previous versions of `pkgconf` the value of the `includedir` variable was returned only once:
```
$ pkgconf --version
2.5.1

$ pkgconf --variable=includedir mumps-dto mumps-zto mumps-sto mumps-cto
C:/msys64/ucrt64/include
```

With a more recent version of `pkgconf`, the value is returned multiple times (presumably once for each pkg-config file?):
```
$ pkgconf --version
3.0.3

$ pkgconf --variable=includedir mumps-dto mumps-zto mumps-sto mumps-cto
C:/msys64/ucrt64/include C:/msys64/ucrt64/include C:/msys64/ucrt64/include C:/msys64/ucrt64/include
```

Is the new behavior intentional? Or is this a regression?

See also: https://github.com/msys2/MINGW-packages/pull/30512

mmuetzel:
If I'm reading the man page correctly, it should return the value of the variable only once (for the first module only):
https://github.com/pkgconf/pkgconf/blob/e87ab86820beed4cd580db94974ed0fa5a869d8e/man/pkgconf.1#L118-L125

https://github.com/pkgconf/pkgconf/blob/e87ab86820beed4cd580db94974ed0fa5a869d8e/man/pkgconf.1#L578-L583

lazka:
bisected to ff07f6b7eb383ce64332562ac664f1cc8ba6f457

possible fix:

```diff
diff --git a/cli/core.c b/cli/core.c
index 7fb8fff..2e1fbb8 100644
--- a/cli/core.c
+++ b/cli/core.c
@@ -1290,6 +1290,10 @@ pkgconf_cli_run(pkgconf_cli_state_t *state, int argc, char *argv[], int last_arg
 	PKGCONF_FOREACH_LIST_ENTRY(deplist.head, node)
 	{
 		pkgconf_dependency_t *dep = node->data;
+
+		if (state->maximum_package_count > 0 && pkgq.length >= state->maximum_package_count)
+			break;
+
 		pkgconf_queue_push_dependency(&pkgq, dep);
 	}
```

kaniini:
Yes, that fix looks correct.

lazka:
I guess this was re-introduced in de0d79afcab53f36dc091ce11f2038691d42c362 ?

[...]

PR #601, opened 2026-09-08
man page: adjust text for recent --variable changes

lazka:
It now prints the variables of all passed modules, not just the first one.

See https://github.com/pkgconf/pkgconf/issues/576#issuecomment-5391344806

----

As a draft since the new behaviour is weird. There is no way to map the output to the input modules if spaces are involved, so I'm having a hard time documenting the form of the output as passing more than one module will result in ambiguous output.

### DIFF
--- a/man/pkgconf.1
+++ b/man/pkgconf.1
@@ -7,7 +7,7 @@
 .\" This software is provided 'as is' and without any warranty, express or
 .\" implied.  In no event shall the authors be liable for any damages arising
 .\" from the use of this software.
-.Dd July 2, 2026
+.Dd September 8, 2026
 .Dt PKGCONF 1
 .Os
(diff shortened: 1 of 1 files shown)


kaniini:
This was reported as supported functionality in original pkg-config.  It prints the variable for each resolved module in sequence.

lazka:
Ah, indeed, weird. Though pkg-config does not add spaces for modules that don't have that variable, while pkgconf does.

```
$ echo "-$(pkg-config --variable=modulesdir openssl libcrypto)-"
-/usr/lib/ossl-modules-
$ echo "-$(pkgconf --variable=modulesdir openssl libcrypto)-"
- /usr/lib/ossl-modules-
```

PR #604, opened 2026-09-11
WIP: Refactor cli, serialization and add libpkgconfcli

illuusio:
Refactor cli code like getopt_long (which is compiled multiple times) and SBOM serialization to `libpkgconfcli`.

This is initiated because of bigger refactoring of SBOM tools and introducing `cdxtool` in future

I've tagged this as WIP as it need bit advise that is this correct refactoring direction and it make big changes. Also should commits be separated differently also should new library be under `cli` and build always static?

Also still learning Meson so it could need some work..


meson.build
```diff
@@ -215,6 +215,27 @@ dep_libpkgconf = declare_dependency(
...
+  libpkgconfcli_sources,
+  c_args: ['-DLIBPKGCONF_EXPORT', build_static],
+  install : true,
+  version : '1.0.0',
+  soversion : '1',
+  link_with : libpkgconf,
+  include_directories : include_directories('libpkgconf'),
+)
+
+# For other projects using libpkgconfig as a subproject
```
moi15moi, line 233: You should add meson.override_dependency()

The wrapdb documentation (https://mesonbuild.com/Adding-new-projects-to-wrapdb.html#creating-the-wrap-contents) says: it is strongly advised in such cases to request that the upstream repository use meson.override_dependency

illuusio: Thank you for reminding with this one. I'm such a N00b with Meson. I've rebased it to first commit.

meson.build
```diff
@@ -502,7 +502,7 @@ if host_machine.system() == 'windows'
   conf_data.set('BOMTOOL_EXE', bomtool_exe.full_path())
   conf_data.set('SPDXTOOL_EXE', spdxtool_exe.full_path())
   conf_data.set('PCCRITIC_EXE', pccritic_exe.full_path())
-  conf_data.set('DLL', libpkgconf.full_path())
+  conf_data.set('DLL', [libpkgconf.full_path(), libpkgconfcli.full_path()])
```
moi15moi: You cannot do that.
You should probably rename `DLL` to `PKGCONF_DLL` and add `PKGCONF_CLI_DLL`.

Don't forget to rename  `DLL` to `PKGCONF_DLL` + add `PKGCONF_CLI_DLL` in this file: https://github.com/pkgconf/pkgconf/blob/4e5a455537845ded1fc11c3cd92bcd4554aa219c/pkgconf.wxs.in#L60-L63

illuusio: Ok this windows build thing is heavily confusing but it didn't fix anything as somehow libpkgconfcli is not build with Windows.

illuusio:
I was thinking that should this be splitted to two PR's like just bare bones `getopt_long` stuff and then serialization it's own as it's much bigger than just to introduce `libpkgconfcli`..

cli/libpkgconfcli/serialize.h
```diff
@@ -0,0 +1,546 @@
...
+
+typedef struct pkgconfcli_serialize_object_list_ {
+	pkgconf_list_t entries;
+} pkgconfcli_serialize_object_list_t;
+
+typedef struct pkgconfcli_serialize_array_ {
+	pkgconf_list_t items;
+} pkgconfcli_serialize_array_t;
+
+bool
```
moi15moi, line 59: The last problem is that these function arent exported, so thats why clang-cl + msvc fail.

To do so, use the macro PKGCONF_API.
Ex: https://github.com/pkgconf/pkgconf/blob/4e5a455537845ded1fc11c3cd92bcd4554aa219c/libpkgconf/libpkgconf.h#L549

